PR #1, opened 2017-06-14
Add the initial module

vmarkovtsev:
It is capable of fetching UASTs


Makefile
```diff
@@ -0,0 +1,35 @@
...
+	@mkdir -p $@
+
+bblfsh/github/com/bblfsh/sdk/protocol/generated_pb2_*.py: \
+        bblfsh/github/com/bblfsh/sdk/protocol github.com/bblfsh/sdk/protocol/generated.proto
+	$(PYTHON) -m grpc.tools.protoc --python_out=bblfsh/github/com/bblfsh/sdk/protocol \
+	    --grpc_python_out=bblfsh/github/com/bblfsh/sdk/protocol \
+	    -I github.com/bblfsh/sdk/protocol -I $(makefile_dir) \
+	    github.com/bblfsh/sdk/protocol/generated.proto
+
+%/__init__.py:
```
juanjux: (non blocking) another option would be to add a first level __init__ with an import hook that knows how to do the rest, but this does the job perfectly too.

bblfsh/channel.py
```diff
@@ -0,0 +1,62 @@
...
+        Initializes a new instance of BblfshClient.
+
+        :param endpoint: The address of the Babelfish server, \
+                         for example "0.0.0.0:9432"
+        :type endpoint: str
+        """
+        self._channel = grpc.insecure_channel(endpoint)
+        self._stub = ProtocolServiceStub(self._channel)
+
+    def fetch_uast(self, file_path, language, contents=None):
```
abeaumont: Note that language could also be None by default for autodetection.

bblfsh/__main__.py
```diff
@@ -0,0 +1,28 @@
...
+
+def setup():
+    parser = argparse.ArgumentParser(
+        description="Query for a UAST to Babelfish and dump it to stdout."
+    )
+    parser.add_argument("-e", "--endpoint", default="0.0.0.0:9432",
+                        help="bblfsh gRPC endpoint.")
+    parser.add_argument("-f", "--file", required=True,
+                        help="File to parse.")
+    parser.add_argument("-l", "--language", required=True,
```
abeaumont: This could be optional for autodetection

bblfsh/test.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @classmethod
+    def tearDownClass(cls):
+        subprocess.check_call("docker rm -f bblfsh".split())
+
+    def setUp(self):
+        self.client = BblfshClient("0.0.0.0:9432")
+
+    def testUAST(self):
+        uast = self.client.fetch_uast(__file__, "Python")
```
juanjux: `uast` in this case is a response object more than a UAST that would be part of it, I'm right? Also should probably check for errors in the response (`response.errors` and/or `response.status`).

bblfsh/test.py
```diff
@@ -0,0 +1,35 @@
...
+import unittest
+
+from bblfsh import BblfshClient
+
+
+class BblfshTests(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        subprocess.check_call(
+            "docker run --privileged -p 9432:9432 --name bblfsh -d "
```
juanjux: Ideally we would have to mock the server to avoid this external dependencies on the test, but this should serve for now.

Makefile
vmarkovtsev: This is very hacky too.

vmarkovtsev:
Update: now it launches the Babelfish server if it is not running (can be disabled).

Makefile
juanjux: It is but done well it could be more generic/automatic for other use cases of this dumb gRPC module.

bblfsh/client.py
```diff
@@ -0,0 +1,64 @@
...
+        Initializes a new instance of BblfshClient.
+
+        :param endpoint: The address of the Babelfish server, \
+                         for example "0.0.0.0:9432"
+        :type endpoint: str
+        """
+        self._channel = grpc.insecure_channel(endpoint)
+        self._stub = ProtocolServiceStub(self._channel)
+
+    def fetch_uast(self, file_path, language=None, contents=None):
```
smola: Please, call this `parse_uast` so that it matches the protocol as much as possible.

smola: Also rename `file_path` to `filename` so it matches too.

bblfsh/client.py
```diff
@@ -0,0 +1,64 @@
...
+
+# The following two insertions fix the broken pb import paths
+sys.path.insert(0, os.path.join(os.path.dirname(__file__),
+                                "github/com/bblfsh/sdk/protocol"))
+sys.path.insert(0, os.path.dirname(__file__))
+from bblfsh.github.com.bblfsh.sdk.protocol.generated_pb2 import ParseUASTRequest
+from bblfsh.github.com.bblfsh.sdk.protocol.generated_pb2_grpc import ProtocolServiceStub
+
+
+class BblfshClient(object):
```
smola: I think this should have a method to close the client and the underlying connection.
Maybe `close()`? I'm not sure that's Python's convention. Also, `__enter__` and `__exit__` so it works with `with`?

vmarkovtsev: There is no close(). Python does it automatically without ResourceWarning-s. Even the Python grpc primer does not close anything: http://www.grpc.io/docs/tutorials/basic/python.html

vmarkovtsev: Bullet proof:

https://github.com/grpc/grpc/blob/master/src/python/grpcio/grpc/_channel.py#L897

vmarkovtsev:
@smola Fixed!

bblfsh/client.py
smola: :+1:

PR #7, opened 2017-07-05
Fix raising the exception while raising the exception

vmarkovtsev:


juanjux:
Nothing to say about the changes in the PR, but some comments and questions about the ensure_bblfsh_is_running function. These are not PR-blocking comments.

- Why sys.stderr.write instead of normal print to stdout?

- The code inside the exception handler could probably go into another function to simplify it a little; I know function calls are expensive in Python and thus the 5 lines-per-function idiom of other languages is not as common, but in this case it doesn't matter much since it's not a performance sensitive function and the code would look nicer removing this long block from the except handler.

- Would be more pythonic using a context manager with the socket object:

```python
with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
```

juanjux:
@smola @abeaumont I don't have write permsions in this project so I'm mentioning you here for review.

juanjux:
@vmarkovtsev it looks like travis is failing, trough the `python -m unittest discover` works on my machine for this project.

vmarkovtsev:
@juanjux writing to stderr because `logging` writes to stderr by default and I am using stdout to write the payload (UAST). Besides, stderr is unbuffered and you do not need to flush.

vmarkovtsev:
@juanjux Regarding the function extraction, I am not using a class here, so it would go to the global names pace and pollute it - I really don't want it.

vmarkovtsev:
Restarting Travis did not help. You updated the image this morning, right?  The first commit was yesterday.

juanjux:
- You can use a __private function, but it's ok in its current form too - just was nitpicking.
- Just so you know, sys.stderr is also line buffered on Python 3 (shouldn't be a problem here since you're finishing lines with a newline) console output and fully buffered for file output.
- Images weren't updated since yesterday, my local test (that worked) was with the latest image.

vmarkovtsev:
Thank you! They appeared to [change the buffering in 3k](https://bugs.python.org/issue13597), didn't know that.

vmarkovtsev:
It hanged on "Babelfish server is up and running." - the socket loop. As if bblfsh server cannot start in Travis env.

juanjux:
Don't worry, I'll take a look later with the Travis docker image, it work locally on my local environment and my testing docker image so it's not blocking IMO, just waiting for another review to merge, maybe Travis is not running with --privileged or something like that.

vmarkovtsev:
```
docker run --privileged -p 9432:9432 --name bblfsh --rm bblfsh/server
Unable to find image 'bblfsh/server:latest' locally
latest: Pulling from bblfsh/server
Status: Downloaded newer image for bblfsh/server:latest
time="2017-07-06T07:54:35Z" level=debug msg="binding to 0.0.0.0:9432" 
time="2017-07-06T07:54:35Z" level=debug msg="initializing runtime at /tmp/bblfsh-runtime" 
invalid image driver format 
time="2017-07-06T07:54:35Z" level=error msg="exiting with error: invalid image driver format " 
The command "docker run --privileged -p 9432:9432 --name bblfsh --rm bblfsh/server" exited with 1.
```

vmarkovtsev:
@juanjux Ta-da

juanjux:
Just for the record of something spoken on Slack, the problem should be fixed on next release of the server docker image.

juanjux:
@smola can you merge this? (I'm still without write permissions on this repo).

PR #16, opened 2017-07-24
feat: add find api

manucorporat:


bblfsh/__main__.py
```diff
@@ -9,26 +9,55 @@ def setup():
     parser = argparse.ArgumentParser(
         description="Query for a UAST to Babelfish and dump it to stdout."
     )
+    parser.add_argument('filename', type=str,
```
juanjux: This change seems to be unrelated with the usage of `libuast`. Was it requested by the machine learning team?

bblfsh/client.py
```diff
@@ -55,8 +56,12 @@ def parse(self, filename, language=None, contents=None, timeout=None,
         request = ParseRequest(filename=os.path.basename(filename),
                                    content=contents,
                                    language=self._scramble_language(language))
-        response = self._stub.Parse(request, timeout=timeout)
-        return response
+        return self._stub.Parse(request, timeout=timeout)
+
+    def parse_uast(self, filename, language=None, contents=None, timeout=None,
```
juanjux: This should be `def parse(` as in `~master`.

bblfsh/__main__.py
```diff
@@ -9,26 +9,55 @@ def setup():
...
 
 def main():
     args = setup()
     if not args.disable_bblfsh_autorun:
         ensure_bblfsh_is_running()
-    client = BblfshClient(args.endpoint)
-    print(client.parse(args.file, args.language))
 
+    client = BblfshClient(args.endpoint)
+    root = client.parse_uast(args.filename, args.language)
```
juanjux: `root = client.parse(`...

bblfsh/test.py
```diff
@@ -8,6 +8,48 @@
...
+
+        node3 = NodeNative("block")
+        node4 = NodeNative("method")
+        node5 = NodeNative("identifier")
+        node5.token = "second"
+        node6 = NodeNative("block")
+        node8 = NodeNative("loop")
+
+        root.add_child(node1)
+
```
juanjux: Spurious whitespace?, probably all the add_child statements would be better on a single logica paragraph (without whitespace between them).

Makefile
```diff
@@ -2,18 +2,26 @@ PYTHON ?= python3
...
+	bblfsh/github/com/bblfsh/sdk/__init__.py \
+	bblfsh/github/com/bblfsh/sdk/uast/__init__.py \
+	bblfsh/github/com/bblfsh/sdk/protocol/__init__.py
+
+libuast:
+	git clone git@github.com:manucorporat/libuast.git libuast
+	cd libuast && cmake . && make && make install
+
+bblfsh/pyuast.py: libuast
+	cd bblfsh && swig -c++ -python -I/usr/local/include uast.i
```
juanjux: So `swig` is needed, at least for `make`. Since the `.i` and `pyuast.py` files are included, couldn't we remove the `all:` target dependency on this one? (but keeping this in the Makefile just in case we want to regenerate `pyuast.py`).

bblfsh/pyuast.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+static PyObject *py_find(PyObject *self, PyObject *args)
+{
+  PyObject *obj = NULL;
+  const char *query = NULL;
+  if (!PyArg_ParseTuple(args, "Os", &obj, &query))
+  {
+    return NULL;
+  }
+  printf("BEFORE FIND!! %p\n", obj);
```
juanjux: These things usually go inside an `#ifdef DEBUG`

juanjux:
Travis still failing ^^^^^^^

bblfsh/__main__.py
```diff
@@ -1,6 +1,8 @@
 import argparse
 import sys
 
+from bblfsh._pyuast import find
```
abeaumont: I don't think pyuast should be a private module, should it?

bblfsh/client.py
```diff
@@ -24,13 +24,14 @@ def __init__(self, endpoint):
                          for example "0.0.0.0:9432"
         :type endpoint: str
         """
+        Node.initialize()
```
abeaumont: I don't think a libuast `Node` should be initialized in a `GRPC` client, why is it this way?

bblfsh/pyuast.cpp
```diff
@@ -0,0 +1,117 @@
...
+    NULL};
+
+PyMODINIT_FUNC PyInit__pyuast(void)
+{
+  return PyModule_Create(&moduledef);
+}
+
+#ifdef __cplusplus
+}
+#endif
```
abeaumont: Newline missing

bblfsh/__main__.py
juanjux: I don't know if there is specified in PEP8 but there is an old convention in Python of using a leading underline for binary modules.

bblfsh/pyuast.c
```diff
@@ -0,0 +1,119 @@
+#include "uast.h"
```
abeaumont: Is there a code style for python modules? If not, I guess we should follow Google C++ code style, which is not been followed here.

bblfsh/pyuast.c
```diff
@@ -0,0 +1,119 @@
...
+#include <stdint.h>
+#include <Python.h>
+
+static const char *read_str(const void *data, const char *prop)
+{
+  PyObject *node = (PyObject *)data;
+  PyObject *attribute = PyObject_GetAttrString(node, prop);
+  const char *a = PyUnicode_AsUTF8(attribute);
```
abeaumont: `return PyUnicode_AsUTF8(attribute)`

bblfsh/pyuast.c
```diff
@@ -0,0 +1,119 @@
...
+      .token = get_token,
+      .children_size = get_children_size,
+      .children = get_child,
+      .roles_size = get_roles_size,
+      .roles = get_role,
+  });
+  return PyModule_Create(&moduledef);
+}
+
+#ifdef __cplusplus
```
abeaumont: This should be removed

setup.py
```diff
@@ -1,5 +1,27 @@
-from setuptools import setup, find_packages
+import sys
+import subprocess
+from setuptools import setup, find_packages, Extension
 
+LIBRARIES = ['xml2']
+SOURCES = ['bblfsh/pyuast.c']
```
abeaumont: These are not constants, they shouldn't be upcase

bblfsh/pyuast.c
juanjux: Not that I know for - or resulted from a quick search. Since @manucorporat probably won't be working on this anymore I'll change it soon.

setup.py
juanjux: I've seen some uppercased module-globals on Python that were not constants but it doesn't seem to be a consensus, PEP8 says that constants are uppercase but doesn't say anything about module (mutable) globals but I think that making them lowercase avoids confusion, yes.

juanjux:
Closing it because I'll reopen this PR from my branch - I already implemented all the suggested changes, but found a crashing bug with some parameters for `find()`, debugging it right now.

juanjux:
To have both PRs linked, the new (unfinished) one is bblfsh/client-python/pull/23

PR #22, opened 2017-08-10
Update protobuf files, bump version

### LINKED ISSUE #21: Numeric role with a specific snippet
With newlines included:

```python


class Repo2nBOW(Repo2Base):
    @property
    def id2vec(self):
        return self._id2vec













```

This produce two instances of role `141` but only with the client-python, it works on the integration tests and other ways to produce the UAST, so probably the bug is in this project.

juanjux:
Fixes #21


juanjux:
Fixes #21

PR #23, opened 2017-09-06
Add libuast, upgrade libuast API, bugfixing

juanjux:
Originally from: bblfsh/client-python/pull/16 moved to my branch to continue the work.

All suggestions in the original PR have been implemented on this branch.

Other changes:

- Added error control which turns an ugly crash into a nice traceback (+ unittest).
- Migrated to `libuast` 0.2.0.
- Migrated to Google's C++ coding style
- Changed autorship to source{d} in the setup.py, updated email.


juanjux:
Looks like the new tests if failing with Python 3.4 which makes CI fails, I'll take a look.

bblfsh/pyuast.c
```diff
@@ -0,0 +1,150 @@
...
+    NULL,
+    NULL,
+    NULL,
+    NULL
+};
+
+PyMODINIT_FUNC PyInit_pyuast(void)
+{
+  NodeIface iface = {
+    .InternalType = GetInternalType,
```
abeaumont: I'd use `InternalType`, etc. instead of `GetInternalType`, etc. so be consistent with code in both `libuast`, its examples, `client-go`, etc.

bblfsh/pyuast.c
```diff
@@ -0,0 +1,150 @@
...
+    NodesFree(nodes);
+    return list;
+  }
+
+  NodesFree(nodes);
+  Py_RETURN_NONE;
+}
+
+static PyMethodDef extension_methods[] = {
+    {"find", PyFind, METH_VARARGS, "Find a node in the UAST"},
```
abeaumont: I'd rename `find` to `filter` for consistency (I changed the name since we're not looking for a node, we're filtering the nodes in the tree that match some predicate.)

juanjux: Since we're not breaking backward compatibility :+1:

juanjux:
Symbols renamed.

PR #31, opened 2017-09-18
move proto files

juanjux:
Seems like I forgot to `git mv` them in the last PR.

Edit: CI it's crashing with a SIGSEGV when calling libuast's `filter`, it's also crashing with the `~master` version (which previously worked) so it must be related to some dependency; I'm investigating the problem.


juanjux:
The crash was because both CI and my tests were using the last Python driver version which changes the UAST.

PR #44, opened 2017-10-17
Add support for libuast 1.3

abeaumont:


bblfsh/pyuast.c
```diff
@@ -14,77 +14,231 @@ static const char *ReadStr(const void *data, const char *prop)
...
+  PyObject *position = PyObject_GetAttrString(node, "start_position");
+  if (!position || position == Py_None) {
+    return 0;
+  }
+
+  PyObject *offset = PyObject_GetAttrString(position, "offset");
+  if (!offset) {
+    return 0;
+  }
+  return PyLong_AsUnsignedLong(offset);
```
juanjux: I think a cast would be more explicit here since `PyLong_AsUnsignedLong` returns an `unsigned long`:

`return (uint32_t)PyLong_AsUnsignedLong(offset);`

Same for the other `uint32_t` functions.

bblfsh/pyuast.c
```diff
@@ -14,77 +14,231 @@ static const char *ReadStr(const void *data, const char *prop)
...
 
-static int ChildrenSize(const void *node)
-{
+static int ChildrenSize(const void *node) {
   return ReadLen(node, "children");
 }
 
-static void *ChildAt(const void *data, int index)
-{
+static void *ChildAt(const void *data, int index) {
```
juanjux: I know it was that way before, but some interface functions use `*node` as parameter and others use `*data`.

abeaumont: Well, since I started with style fixes I should also fix that =)

bblfsh/pyuast.c
abeaumont: I didn't do it since I thought function signature was explicit enough in this case. I can add it though.

abeaumont:
Have simplified the code, made it more consistent, and added review feedback.

juanjux:
I like it, I'll adapt those remade helper functions (`Attributte, String, Size`, etc) for the Scala client too. 

Can it be merged? (the CI is working).

abeaumont:
Not yet, current testsuite doesn't cover much about uast filter, I want to improve that before merging.

abeaumont:
@juanjux added the tests and now there're random segfaults, even if all the tests pass individually. I think current bindings have some kind of memory handling problem.

abeaumont:
Found and fixed the reference counter problem, added some more tests, and now everything seems to work properly. I'd say this is ready to be merged :)

juanjux:
@alfredo, before your last message I was about to tell you that the Scala part is working well now,  but I want to torture test it a little more and maybe do a memory profiling with Valgrind to see if there is any leak (at least with the Python client that should use refcounting, the JVM is less deterministic and thus harder to evaluate for leaks). Anyway I'm tagging so people can start using the new filters.

Issue #29, opened 2017-09-08
Does bblfsh.filter should return list (instead of iterator)?


zurk:
It is a suggestion of enhancement. 
Sometimes I don't need a full list, and it is a more pythonic way to have an iterator instead of the list. 

To discuss.

abeaumont:
I think it's a good idea and should be easy to implement.

juanjux:
Agree.

juanjux:
Fixed by #46.

PR #46, opened 2017-10-23
[RFC] Feature/iterator result

juanjux:
From #29.

Changes `filter()` to return an iterator object instead of a list. This goes in line with what most Python 3.x APIs do. It also breaks any code doing a `len(result)` or `result[index]` which would have to be changed to `len(list(result))`, `list(result)[0]` or `any(result)` to check if there is any element. 

This was agreed as a good thing on the linked issue but I think we should get feedback from other Python client users before merging.


juanjux:
Since there wasn't any feedback on the Slack channel I tacitly understand nobody is against this change and thus will merge.

PR #52, opened 2017-10-31
Get protofiles from SDK release. Remove Makefile.

### LINKED ISSUE #36: Remove the `Makefile`, make `setup.py` retrieve the `.pb` files from release
Doesn't make sense to have a `Makefile` when we have a `setup.py` file, also, we could remove the .pb files that were added to avoid a dependency on `protoc` from the `pip` package.

- [ ] Move the functionality of the `Makefile` to the `setup.py`.
- [ ] A `--release` switch should be added to `setup.py` to generate the `.pb` files.
- [ ] The travis build should install `protoc` and call `--release` before the `pip install`.
- [ ] The normal build called by `pip install` should check if the pb files are generated and if not, retrieve the right release zip from github an extract the files from there before install.

juanjux:
Fixes #36.

Also compiles protobuf from sources on the Travis build since there aren't packages from it for Travis' Ubuntu 14.04. I'm not totally satisfied with it since it takes now 24 minutes to complete the build, I'll investigate options to upload a precompiled package somewhere and retrieve it on build.


setup.py
```diff
@@ -1,54 +1,153 @@
...
-        "Intended Audience :: Developers",
-        "License :: OSI Approved :: Apache Software License",
-        "Operating System :: POSIX",
-        "Programming Language :: Python :: 2.7",
-        "Programming Language :: Python :: 3.4",
-        "Programming Language :: Python :: 3.5",
-        "Programming Language :: Python :: 3.6",
-        "Topic :: Software Development :: Libraries"
+LIBUAST_VERSION = "v1.3.0"
+SDK_MAJOR = "v1"
```
abeaumont: minor: this could be extracted from `SDK_VERSION` and remove the possibility of setting incompatible versions.

juanjux: True!

juanjux:
Yes, I'll look today into compiling it in an Ubuntu 14.04 VM or container and uploading it to google drive.

juanjux:
@abeaumont fixed the compiling kludge, I compiled a deb package for 14.04 and since it seem there isn't an easy/elegant way to download a file non interactively from gdriver, I uploaded the deb in a release in this same repository.

PR #56, opened 2017-11-03
libuast: Bump to 1.4.1 and improve error handling

abeaumont:
Signed-off-by: Alfredo Beaumont <alfredo.beaumont@gmail.com>


juanjux:
CI fails because of the ancient GCC in Travis. Add `-std=gnu99` to the flags, which I don't know exactly how its done for Python packages, or declare the int outside the for loop.

abeaumont:
I've explicitly removed `c99` because it doesn't include `strndup`... I'll use `gnu99` instead

Issue #60, opened 2017-11-24
SIGSEGV on filtering //@roleLiteral


vmarkovtsev:
I am using the most recent bblfsh client, exec the following code:

```
import bblfsh
client = bblfsh.BblfshClient("0.0.0.0:9432")
uast = client.parse("/path/to/src-d/gemini/gemini/python/setup.py").uast
list(bblfsh.filter(uast, "//@roleLiteral"))
```

I get SIGSEGV. `gdb bt`:

```
#0  PyFilter (self=<optimized out>, args=<optimized out>) at bblfsh/pyuast.c:168
#1  0x00000000004e10ef in PyCFunction_Call () at ../Objects/methodobject.c:109
#2  0x00000000005240b4 in call_function (oparg=<optimized out>, pp_stack=0x7fffffffbde0) at ../Python/ceval.c:4705
#3  PyEval_EvalFrameEx () at ../Python/ceval.c:3236
#4  0x000000000052cf19 in _PyEval_EvalCodeWithName () at ../Python/ceval.c:4018
#5  0x000000000052dbcf in PyEval_EvalCodeEx () at ../Python/ceval.c:4039
#6  PyEval_EvalCode (co=<optimized out>, globals=<optimized out>, locals=<optimized out>) at ../Python/ceval.c:777
#7  0x0000000000543075 in builtin_exec_impl.isra.11 (locals=
[...]

PR #64, opened 2017-12-06
Re-raise UTF8 grpc encoding error in a clearer way

### LINKED ISSUE #61: UnicodeDecodeError in client.parse()
I try to run bblfsh from python for this file: https://github.com/INRIA/spoon/blob/master/src/test/resources/noclasspath/IsoEncoding.java
As you can see it is the file with non-standard characters.

If I run
```python
import bblfsh

client = bblfsh.BblfshClient("0.0.0.0:9234")
res = client.parse("IsoEncoding.java")
```
I get
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/google/protobuf/internal/type_checkers.py", line 182, in CheckValue
    proposed_value = proposed_value.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 69: invalid continuation byte

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Applications/PyCharm.app/Contents/helpers/pydev/pydevd.py", line 1599, in <module>
    globals = debugger.run(setup['file'], None, None, is_module)
  File "/Applications/PyCharm.app/Contents/helpers/pydev/pydevd.py", line 1026, in run
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/Applications/PyCharm.app/Contents/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/Users/k/sourced/ast2vec/how_to_use_ast2vec.py", line 4, in <module>
    res = client.parse("../bad_file/IsoEncoding.java")
  File "/usr/local/lib/python3.6/site-packages/bblfsh/client.py", line 56, in parse
  File "/usr/local/lib/python3.6/site-packages/google/protobuf/internal/python_message.py", line 526, in init
    setattr(self, field_name, field_value)
  File "/usr/local/lib/python3.6/site-packages/google/protobuf/internal/python_message.py", line 662, in field_setter
    new_value = type_checker.CheckValue(new_value)
  File "/usr/local/lib/python3.6/site-packages/google/protobuf/internal/type_checkers.py", line 187, in CheckValue
    (proposed_value))
ValueError: b'package my.pack;\n\npublic class IsoEncoding {\n\n    private String entr\xe9e;\n    public IsoEncoding(String maCha\xeene) {\n        this.entr\xe9e = maCha\xeene;\n    }\n\n    private String r\xe9cup\xe9rerUneEntr\xe9e() {\n        return this.entr\xe9e;\n    }\n}' has type bytes, but isn't valid UTF-8 encoding. Non-UTF-8 strings must be converted to unicode objects before being added.
```

Is it expected behavior for non-unicode files?
If so I think it is better to provide better error message from bblfsh client.

juanjux:
Fixes #61

It branches from #63 since CI won't work without it (because there is no bblfsh/server anymore), but disables the test for #60 since it would fail until libuast is tagged (thus this will show the commits of the previous PR too).

The non carried-over changes from this PR are [here](https://github.com/bblfsh/client-python/pull/64/files#diff-fbbd9410dbf75b0a91dbb45c2b1d66b8R56) and [here](https://github.com/bblfsh/client-python/pull/64/files#diff-45858e403cdd5d4a298121d17a9ddaafR39).

I'll keep the other PR open so I remember to enable that test once libuast has been tagged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ## client-python [![Build Status](https://travis-ci.org/bblfsh/client-python.svg?branch=master)](https://travis-ci.org/bblfsh/client-python) [![PyPI](https://img.shields.io/pypi/v/bblfsh.svg)](https://pypi.python.org/pypi/bblfsh)
 
 [Babelfish](https://doc.bblf.sh) Python client library provides functionality to both
-connect to the Babelfish server to parse code
+connect to the Babelfish bblfshd to parse code
 (obtaining an [UAST](https://doc.bblf.sh/uast/specification.html) as a result)
 and to analyse UASTs with the functionality provided by [libuast](https://github.com/bblfsh/libuast).
 
@@ -37,10 +37,11 @@ A small example of how to parse a Python file and extract the import declaration
 If you don't have a bblfsh server running you can execute it using the following command:
 
 ```sh
-docker run --privileged --rm -it -p 9432:9432 --name bblfsh bblfsh/server
+docker run --privileged --rm -it -p 9432:9432 -v bblfsh_cache:/var/lib/bblfshd --name bblfshd bblfsh/bblfshd
+docker exec -it bblfshd bblfshctl driver install python bblfsh/python-driver:latest
 ```
 
-Please, read the [getting started](https://doc.bblf.sh/user/getting-started.html) guide to learn more about how to use and deploy a bblfsh server.
+Please, read the [getting started](https://doc.bblf.sh/user/getting-started.html) guide to learn more about how to use and deploy a bblfshd.
 
 ```python
 from bblfsh import BblfshClient, filter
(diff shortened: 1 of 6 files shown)


juanjux:
Mmm I only remember we just said to make the exception more readable and that's why we added it to the planning. 

There are several reasonable options as I see it:

1. Keep the original (pretty unreadable for people not familiar with Python 3) traceback.
2. Keep the original exception type (ValueError, usually discouraged to use because it's too generic) but change the message and discard the original traceback. So we're really raising a new ValueError but with a different message (raising the original ValueError would keep the unreadable traceback).
3. Wrap into a new exception, with a new message, and discarding the previous traceback.
4. Same as 3) but keeping the original traceback.

I took 3), 2) could be reasonable too. 4) would be as unreadable as 1. What would you prefer in this case?

abeaumont:
If I don't remember it wrong, what we discussed was that:
- It made no sense to wrap exceptions of inner components (i. e. why this one and no the rest of exceptions grpc could throw?)
- The client should prevent this exception from raising because it should make sure that code is correctly encoded

juanjux:
Second point makes sense. I'll implement the test for non UTF8 content before sending to gRPC. Closing this PR.

PR #66, opened 2017-12-22
Add libuast iterators support

### LINKED ISSUE #62: Add support for iterators

juanjux:
Fixes #62.

Note: CI will fail until the libuast functions' PR have been merged and tagged (I used 1.6.0 in this PR, but could update) because of the change of `int` to `size_t` in nodeiface `**Size` interface functions.


README.md
```diff
@@ -44,13 +44,18 @@ docker exec -it bblfshd bblfshctl driver install python bblfsh/python-driver:lat
...
+client = bblfsh.BblfshClient("0.0.0.0:9432")
 uast = client.parse("/path/to/file.py").uast
 print(uast)
 # "filter' allows you to use XPath queries to filter on result nodes:
-print(filter(uast, "//Import[@roleImport and @roleDeclaration]//alias")
+print(bblfsh.filter(uast, "//Import[@roleImport and @roleDeclaration]//alias")
+
+# You can also iterate on several tree iteration orders:
+it = bblfsh.iterator(uast, bblfsh.TreeOrder.pre_order)
+for node in it:
```
abeaumont: I think something like this would be more idiomatic, not sure how more difficult this would be to implement:

```
for node in uast.iter(bblfsh.TreeOrder.PRE_ORDER):
```

bblfsh/pyuast.c
```diff
@@ -146,10 +146,111 @@ static Uast *ctx;
...
+  PyUastIter *pyIt = PyObject_New(PyUastIter, &PyUastIterType);
+  if (!pyIt)
+    return NULL;
+
+  if (!PyObject_Init((PyObject *)pyIt, &PyUastIterType)) {
+    Py_DECREF(pyIt);
+    return NULL;
+  }
+
+  pyIt->iter = UastIteratorNew(ctx, node, (TreeOrder)order);
```
abeaumont: Check for null missing either here or in the caller code

README.md
juanjux: I checked a while ago (when I did something similar for the Scala client) and it was not trivial, so I would leave that for another PR.

bblfsh/__init__.py
```diff
@@ -1,11 +1,15 @@
 from bblfsh.client import BblfshClient
-from bblfsh.pyuast import filter
+from bblfsh.pyuast import filter, iterator
 from bblfsh.aliases import *
 
+class TreeOrder:
+    pre_order = 0
```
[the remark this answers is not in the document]
juanjux: Yes, that's right for separate constants, but it's pretty common that enum-like constructs are  done with classes like here, named tuples or the enum in Python 3.4 (but I didn't want to use it yet to break compatibility with older 3.x versions). Python 3.7 will introduce another one ("data classes"). It's not very important anyway, but I think the class with static members option keep things tidier than separate constants.

juanjux:
Yes, my idea was to add the filtering functions in another PR after the Scala and Go clients have the iterators too.

bblfsh/__init__.py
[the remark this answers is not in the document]
abeaumont: Well, I was mostly referring to the _written in all capital letters with underscores separating words. Examples include MAX_OVERFLOW and TOTAL._ part ;)

bblfsh/pyuast.c
```diff
@@ -237,6 +237,9 @@ static PyObject *PyUastIter_new(PyObject *self, PyObject *args)
   }
 
   pyIt->iter = UastIteratorNew(ctx, node, (TreeOrder)order);
+  if (!pyIt->iter)
```
abeaumont: I guess a `Py_DECREF(pyIt);` is also needed here

bblfsh/__init__.py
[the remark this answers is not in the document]
juanjux: :man_facepalming: yes, they should be uppercase, I'll do it.

PR #69, opened 2018-01-02
docs: Update installation instructions

abeaumont:
Signed-off-by: Alfredo Beaumont <alfredo.beaumont@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ pip install bblfsh
 ```bash
 git clone https://github.com/bblfsh/client-python.git
 cd client-python
-make install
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,7 +18,7 @@ pip install bblfsh
 ```bash
 git clone https://github.com/bblfsh/client-python.git
 cd client-python
-make install
+python setup.py install
```
juanjux: I think the "more modern way" would be `pip install .`  but I just checked and this also shows it in pip package list so it's Ok.

PR #75, opened 2018-01-31
Fix memory leaks (thanks @amlweems)

juanjux:
(Branched from amlweems:fix-memory-leak)

- Apply the memory leaks [detected and fixed by @amlweems](https://github.com/bblfsh/client-python/pull/74) with an aditional one for the `String` helper functions that allows the fix for `ItemAt` to work.

- Test (also from @amlweems branch) to check that memory doesn't increase after 100 `filter()` calls. 

### DIFF
--- a/bblfsh/pyuast.c
+++ b/bblfsh/pyuast.c
@@ -2,8 +2,16 @@
 
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdlib.h>
+#include <string.h>
 
 #include "uast.h"
+
+// Used to store references to the Pyobjects instanced in String() and
+// ItemAt() methods. Those can't be DECREF'ed to 0 because libuast uses the
+// so we pass ownership to these lists and free them at the end of filter()
+static PyObject *stringAllocsList;
+static PyObject *itemAtAllocsList;
 
 static PyObject *Attribute(const void *node, const char *prop) {
   PyObject *n = (PyObject *)node;
@@ -20,14 +28,21 @@ static const char *String(const void *node, const char *prop) {
   PyObject *o = Attribute(node, prop);
   if (o != NULL) {
     retval = PyUnicode_AsUTF8(o);
+    PyList_Append(stringAllocsList, o);
     Py_DECREF(o);
   }
   return retval;
 }
 
 static size_t Size(const void *node, const char *prop) {
+  size_t retval = 0;
   PyObject *o = Attribute(node, prop);
-  return o ? PySequence_Size(o) : 0;
+  if (o != NULL) {
+    retval = PySequence_Size(o);
+    Py_DECREF(o);
+  }
+
+  return retval;
 }
 
 static PyObject *ItemAt(PyObject *object, int index) {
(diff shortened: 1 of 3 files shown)


juanjux:
cc @amlweems

amlweems:
Nice! Testing with `pympler` again, this fixes the major memory leak. And the objects that are still around don't scale with the number of iterations I do against `filter()`.

```
# 5000 iterations of filter()
        types |   # objects |   total size
============= | =========== | ============
  <class list |        5056 |    555.20 KB
   <class str |        5045 |    438.51 KB
   <class int |        1003 |     43.10 KB
...

# 10000 iterations of filter()
        types |   # objects |   total size
============= | =========== | ============
  <class list |        5056 |    555.20 KB
   <class str |        5045 |    438.51 KB
   <class int |        1003 |     43.10 KB
...
```

I tried running with cpython and valgrind. I get a lot of "possibly lost", but no worrying "definitely lost". I'm not super confident in these results, but I am very happy with the object counts above. [Full output attached.](https://github.com/bblfsh/client-python/files/1682043/leaks.txt)

```
# cpython build conf
./configure --without-pymalloc --with-pydebug --with-valgrind --prefix=`pwd`/build

# valgrind command
valgrind --tool=memcheck --suppressions=/tmp/cpython/Misc/valgrind-python.supp --log-file=leaks.txt --leak-check=full python -m unittest bblfsh.test.BblfshTests.testManyFilters
```

```
==21838== LEAK SUMMARY:
==21838==    definitely lost: 40 bytes in 1 blocks
==21838==    indirectly lost: 0 bytes in 0 blocks
==21838==      possibly lost: 6,314,378 bytes in 37,480 blocks
==21838==    still reachable: 193,285 bytes in 310 blocks
==21838==         suppressed: 0 bytes in 0 blocks
```

juanjux:
@vmarkovtsev that would be certainly great, I used a simpler tool that doesn't Python to recompile called memleak, I'm pretty sure current leaks are fixed since I increased locally the iterations in the memory test to 100.000 and the memory didn't increase, but I'll give a run with Valgrind when I've a little time.

On the other side, Travis is failing with a core dump while works perfectly on my machine :/

vmarkovtsev:
@amlweems This leak summary gives a solid confidence that all the leaks are fixed, awesome!

juanjux:
Found the reason why it works on my machine(TM) and not in Travis... when testing for leaks, I set the environment variable "PYTHONMALLOC" to "malloc" since that allows the `memleak` command to find the leaks. With that environment variable set, it works, without it, it segfaults. Looks like I'm for another debugging session :-1:

juanjux:
Looks like the problem comes from mixing malloc with pymalloc in the same extension module. This is not a problem for shared libraries, but since libuast is compiled with the extension module it causes problems. The solutions I'm evaluating are requiring and using `libuast.so` or adding a function to libuast to pass the alloc/free functions to use.

juanjux:
Ok, the pointer-between-mallocs problem is hellish and using the `libuast.so` crashed just the same, so I did a workaround that while not perfect will avoid the leak: I store the problematic PyObjects in a PyList  and free (DECREF) it at the end of Python's `filter()` wrapper when libuast is not touching them anymore for sure.

vmarkovtsev:
Amazing detective work @juanjux !

PR #88, opened 2018-03-21
Change position iterator unittest to match the expected ordering of the fixed libuast

juanjux:
Requires: bblfsh/libuast/pull/79


juanjux:
Should work now.

PR #90, opened 2018-03-28
Doc: update README, add required native dependency

bzz:
Adding to dependency section of README as `pip install` depends on it:
  - `curl` i.e https://github.com/bblfsh/client-python/blob/0037d762563ab49b3daac8a7577f7103a5628fc6/setup.py#L72
  - `g++` https://github.com/bblfsh/client-python/blob/0037d762563ab49b3daac8a7577f7103a5628fc6/setup.py#L17

Otherwise `pip install bblfsh` fails on a system, if those native packages not present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,13 @@ python setup.py install
 
 ### Dependencies
 
-You need to install `libxml2` and its header files. You also will need a `curl` cli tool. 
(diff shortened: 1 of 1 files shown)


juanjux:
Thanks for the PR!

PR #99, opened 2018-04-23
[WIP] travis: support osx and python 3.4

smola:
Signed-off-by: Santiago M. Mola <santi@mola.io>


smola:
No Python support on OS X for Travis CI at the moment: https://github.com/travis-ci/travis-ci/issues/2312
Possible workarounds exist (they are discussed in the issue above), such as https://github.com/MacPython/terryfy

juanjux:
@smola so is this ready to merge or a WIP? (I see that CI was failing, I guess for the reason you wrote).

smola:
@juanjux WIP, I need to figure out the right way to setup Python on macOS. it should be using official Python.org installers, which should be fast and also it is the only way to generate macOS wheels.

smola:
Closing, this will just not work in its current form.

PR #107, opened 2018-07-25
Rework setup.py to be portable

vmarkovtsev:
- Replace calls to coreutils with equivalent Python APIs
- os.path.join


juanjux:
@vmarkovtsev  before you go mad with it: the CI error doesn't seem to be related with your PR. I'm getting the same error with `~master`, looking into it right now.

vmarkovtsev:
@juanjux Fixed

PR #112, opened 2018-07-27
Add the SupportedLanguages protocol call

### LINKED ISSUE #98: [feature request] Add support to query supported languages (with installed drivers)
Similar to https://github.com/bblfsh/client-scala/issues/68 it is really useful to have this feature in Python client.
Now we (ML team) have a hardcoded list of languages:  https://github.com/src-d/ml/blob/master/sourced/ml/cmd_entries/args.py#L36-L40 and we should update it each time.

juanjux:
Fixes #98. Also updates an unittest to make it compatible with v1 and v2 branches and add the recently released Python 3.7 to the CI.

cc @zurk

### DIFF
--- a/bblfsh/client.py
+++ b/bblfsh/client.py
@@ -3,7 +3,8 @@ import sys
 
 import grpc
 
-from bblfsh.aliases import ParseRequest, NativeParseRequest, VersionRequest, ProtocolServiceStub
+from bblfsh.aliases import (ParseRequest, NativeParseRequest, VersionRequest,
+        ProtocolServiceStub, SupportedLanguagesRequest, SupportedLanguagesResponse)
 from bblfsh.sdkversion import VERSION
 
 # The following two insertions fix the broken pb import paths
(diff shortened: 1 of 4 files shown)


bblfsh/client.py
```diff
@@ -99,6 +100,14 @@ def native_parse(self, filename, language=None, contents=None, timeout=None):
                                language=self._scramble_language(language))
         return self._stub.NativeParse(request, timeout=timeout)
 
+    def supported_languages(self):
+        sup_response = self._stub.SupportedLanguages(SupportedLanguagesRequest())
+
+        try:
+            return sup_response.ListFields()[1][1]
```
dennwc: Maybe a dumb question, but why it's an index access instead of field access?

juanjux: Brainfart :) I was testing it the response could be accessed with that method as a normal python indexed list and the objects and dictionaries and forgot to restore the access to languages.

I'll change it in the next commit + a unittest.

Issue #122, opened 2018-08-29
Different Node classes in the same package


vmarkovtsev:
```python
import importlib
import bblfsh

m = importlib.import_module('bblfsh.gopkg.in.bblfsh.sdk.v1.protocol.generated_pb2_grpc')
print(id(m.generated__pb2.gopkg_dot_in_dot_bblfsh_dot_sdk_dot_v1_dot_uast_dot_generated__pb2.Node))
print(id(bblfsh.Node))
m = importlib.import_module('bblfsh.gopkg.in.bblfsh.sdk.v1.uast.generated_pb2')
print(id(m.Node))
```

The first printed number is different from the rest. That's because `bblfsh/gopkg/in/bblfsh/sdk/v1/protocol/generated_pb2_grpc.py` does

```
import generated_pb2 as generated__pb2
```

As soon as the module is imported from the different PYTHONPATH, the classes get screwed and although they are really the same, they are different. Hence `isinstance()` returns False on 3.7, etc.

Suggested fix: remove all the tricks with PYTHONPATH and make all the imports absolute everywhere and starting with `bblfsh.`.

vmarkovtsev:
I am on it.

vmarkovtsev:
Fixed long time ago

PR #123, opened 2018-09-04
Fix PB imports (#122)

vmarkovtsev:
So I had to do several things:

1. Place pb files inside proto/bblfsh
2. Patch their imports to start with bblfsh
3. Cast necromancery spells at the generated grpc stubs. The grpc plugin is crazy and ignores ".in." in imports, it also does not respect the directory structure.

So now we have absolute imports starting with "bblfsh" **everywhere** and no classes are duplicated.


vmarkovtsev:
@bzz Alarma! When this is merged, the way Python bindings in the Python (ML) SDK are built must be changed. I will work on the fix and push it to style-analyzer today.

vmarkovtsev:
Ah it is too complex to fix the downstream projects, I have to move protos from "bblfsh" to the root.

PR #124, opened 2018-09-04
Fix PB imports (#122)

vmarkovtsev:
Changes:

1. Place pb files inside proto
2. Fix the generated grpc stubs. The grpc plugin is crazy and ignores ".in." in imports, it also does not respect the directory structure.
3. Prepend "bblfsh" to all the imports in the bindings
4. Remove the sys.path hacks
5. Add `isinstance` test

So now we have absolute imports starting with "bblfsh" **everywhere** and no classes are duplicated.

Signed-off-by: Vadim Markovtsev <vadim@sourced.tech>


vmarkovtsev:
@bzz This is the proper fix. No changes are required for the downstream projects.

vmarkovtsev:
@dennwc Can you please merge and publish a new release on PyPi.

PR #128, opened 2018-10-04
Python client v3 (UASTv2)

dennwc:
Work-in-progress implementation of Python client v3 for the new libuast.

**TODO:**
- [x] Proper memory management
- [x] Fix install scripts
- [x] Update gRPC client to use v1 and v2 protocols
- [x] Test that the static linking works in Windows and macos and include static files in the libuast package.
- [x] Tests
- [x] Fix some TO-DOs in the code about potential memory problems.
- [x] Update `README.md`
- [x] Enable failing test `testFilterToken` once the SDK problem has been fixed.
- [x] Add `Node` compatibility layer.

Signed-off-by: Denys Smirnov <denys@sourced.tech>


juanjux:
Install script should be fixed.

juanjux:
@dennwc PTAL at the last commit. Should work on Linux (works on my machine and my Docker image) but as we talked on Slack it'll require the libuast package to include static libraries and it's untested on Windows and Darwin. It also have `GET_LIBUAST` to False to ease testing this, should be changed before merging.

README.md
```diff
@@ -48,16 +48,18 @@ Please, read the [getting started](https://doc.bblf.sh/using-babelfish/getting-s
...
+it = uast.filter("//Import[@role='Import' and @role='Declaration']//alias")
+for node in it:
+    print(node.load())
 
-# filter\_[bool|string|number] must be used when using XPath functions returning
-# these types:
-print(bblfsh.filter_bool(uast, "boolean(//*[@strtOffset or @endOffset])"))
-print(bblfsh.filter_string(uast, "name(//*[1])"))
-print(bblfsh.filter_number(uast, "count(//*)"))
+# filter must be used when using XPath functions returning these types:
```
juanjux: The comment about the types is spurious if filter now works with different types.

juanjux: One worry I have with this version of the client is that it seems to be backward incompatible with existing code (as these functions and the returned value show) which is not something I'm aware was advertised since the rewrite of libuast should have keep the clients APIs the same (or at least I'm pretty sure most users of the Python client are not aware of this).

dennwc: But it will be incompatible in any case, since:
- `Node` does not exist and I don't think we should emulate it.
- `load()` was not used in the old code, but we need it for better performance.
- `filter` returned a node or an array, but now it will return an iterator.

We can still emulate `filter_*` if necessary. But I think a generic `filter_one` or `filter_val` will be better, since Python is not strictly typed, and it's nice to enter the query and get a value back whatever the type is.

dennwc: Tagging @vmarkovtsev to share opinions about backward compatibility.

juanjux: filter() returned a Python iterator already, not arrays or single nodes. I'm not saying we should not improve, just that we should not break backwards compatibility without checking for impact before with users and management.

juanjux: On the leads meeting @mcuadros didn't saw any problem breaking backward compatibility on v3 so if @vmarkovtsev doesn't have a problem with that either I think we can go ahead with this.

juanjux: Having solved the compatibility stuff and returning to the other part, I still think that a single `filter` function returning different types is worse than a function named for every type (because it's super unidiomatic and forces the user to check the type on every call and doesn't check if the user is getting the wrong/unexpected type).

dennwc: It relies on the returned type of the query here. `name(*)` always returns a string, for example.

juanjux:
@dennwc I would prefer if you debugged the memory problems that you indicated by the TO-DOs since you're more familiar with the new libuast and the C++ binding. I could write the tests once I finish the C++ driver.

dennwc:
@juanjux Should be good from memory management point of view. Let's add tests and see if it breaks there :grin:

juanjux:
Very nice! I'll work on the tests now.

juanjux:
@dennwc unittests migrated. There is a failing test that I've left commented out, `testFiltertoken`, PTAL, this xpath query:

```python
it = ctx.filter("//*[@token='else']/@token")
```

AFAIK it should return the `@token` property value (`else`) but it returns the full dictionary containing it.

juanjux:
Uncommented the failing test so we don't merge by accident.

dennwc:
@juanjux Checked the test case, it looks like it's a bug either in SDK or XPath library. Opened https://github.com/bblfsh/sdk/issues/340 for it.

So if everything else works here and @vmarkovtsev is happy with Python API, we should comment this case out (until SDK is fixed) and finally release the client.

juanjux:
It's a pretty important case since a lot of the real xpath usage involves extracting tokens. I'll take a look at the SDK and the Xpath lib tomorrow, maybe it's something easy. If it isn't, we could implement a workaround while this is fixed.

dennwc:
Agree, but still, it's not related to the client - it just dumps whatever `libuast` (SDK) returns.

But yeah, makes sense to try to fix libuast before tagging a release. Or tag an RC to get some feedback on the API first.

juanjux:
@mcuadros already approved of breaking changes in the API on a leads meeting so I'm commenting that test and merging if everybody here agrees.

bzz:
@juanjux awesome! Please give some time to try it out on macOS again later today

bblfsh/pyuast.cc
```diff
@@ -1,272 +1,760 @@
...
+    } catch (const std::exception& e) {
+        PyErr_SetString(PyExc_RuntimeError, e.what());
+    }
+    return it;
+}
 
-  static uint32_t StartCol(const void *node) {
-    return PositionValue(node, "start_position", "col");
-  }
+// PythonContextExt_filter serializes UAST.
```
bzz: But the function is `PythonContextExt_encode` 🤔

bblfsh/pyuast.cc
```diff
@@ -1,272 +1,760 @@
...
-    retval = PySequence_Size(o);
-    Py_DECREF(o);
-  }
+class ContextExt;
 
-  return retval;
-}
+typedef struct {
+  PyObject_HEAD
+  ContextExt    *ctx;
```
bzz: Probably a nit pick, but is there any particular reason for this formatting?

bblfsh/pyuast.cc
```diff
@@ -1,272 +1,760 @@
...
+    //return PyMemoryView_FromMemory((char*)(buf.ptr), buf.size, PyBUF_READ);
 }
 
-// WARNING: calls to AttributeValue MUST Py_DECREF the returned value once
-// used (or add it to the memtracker)
-static PyObject *AttributeValue(const void *node, const char *prop) {
-  PyObject *a = Attribute(node, prop);
-  return a && a != Py_None ? a : NULL;
+/*
+static bool checkError(const Uast* ctx) {
```
bzz: Is there a reason to keep this commented?

bblfsh/pyuast.cc
```diff
@@ -1,272 +1,760 @@
...
-// WARNING: calls to Attribute MUST Py_DECREF the returned value once
-// used (or add it to the memtracker)
-static PyObject *Attribute(const void *node, const char *prop) {
-  PyObject *n = (PyObject *)node;
-  return PyObject_GetAttrString(n, prop);
+PyObject* asPyBuffer(uast::Buffer buf) {
+    PyObject* arr = PyByteArray_FromStringAndSize((const char*)(buf.ptr), buf.size);
+    free(buf.ptr);
+    return arr;
+    //return PyMemoryView_FromMemory((char*)(buf.ptr), buf.size, PyBUF_READ);
```
bzz: Is there a reason why we want to keep this comment here?

It that is for dev documentation purpose - e.g "how to do debuging", it's cool to keep it but a comment why would really help to understand that for a future reader.

bblfsh/pyuast.cc
```diff
@@ -1,272 +1,760 @@
...
-    return PositionValue(node, "end_position", "offset");
-  }
+extern "C"
+{
+    static PyTypeObject PythonContextExtType = {
+      PyVarObject_HEAD_INIT(nullptr, 0)
+      "pyuast.ContextExt",               // tp_name
+      sizeof(PythonContextExt),                 // tp_basicsize
+      0,                              // tp_itemsize
+      PythonContextExt_dealloc,                 // tp_dealloc
```
bzz: Here, above and in some other `PyTypeObject`s formatting seems to be a bit off.

bblfsh/pyuast.cc
```diff
@@ -1,272 +1,760 @@
...
+static PyObject *PyUastIterExt_next(PyObject *self) {
+  auto it = (PyUastIterExt *)self;
 
-static size_t PropertiesSize(const void *node) {
-  size_t retval = 0;
-  PyObject *properties = AttributeValue(node, "properties");
-  if (properties) {
-    retval = PyMapping_Size(properties);
-    Py_DECREF(properties);
+  // TODO: check errors
```
bzz: Is this still relevant and if it is, could it affect the correctness of the result?

bblfsh/pyuast.cc
```diff
@@ -1,272 +1,760 @@
...
+        }
+
+        auto node = (PyNodeExt*)obj;
+        return node->handle;
+    }
+
+    PyObject* newIter(uast::Iterator<NodeHandle> *it, bool freeCtx){
+        PyUastIterExt *pyIt = PyObject_New(PyUastIterExt, &PyUastIterExtType);
+        if (!pyIt)
+          return nullptr;
```
bzz: According to [C++ styleguide #Conditionals](https://google.github.io/styleguide/cppguide.html#Conditionals)

> In general, curly braces are not required for single-line statements, but they are allowed if you like them;

So it would be nice to be consistent and either have this (and all other cases like this in this file) as one-liner (as above), or add braces [as below](https://github.com/bblfsh/client-python/pull/128/files#diff-8bd44ca8121fa182cb0123e19483c6f4R1068).

bblfsh/pyuast.cc
juanjux: Fixed in next commit.

bblfsh/pyuast.cc
juanjux: Fixed in next commit.

bblfsh/pyuast.cc
juanjux: Fixed in next commit.

bblfsh/pyuast.cc
juanjux: Fixed in next commit. I've moved to use the one-liner version that was already used sparingly since that's less verbose and also avoids errors accidentally trying to add more code to an unbraced if, except in cases where the line length would be too high.

juanjux:
@bzz fixed some of the feedback, thanks! I'll leave some for @dennwc since that commented stuff seems to be probably related with the memory debugging that he did.

bzz:
@juanjux awesome, thank you!

Two questions:

> @mcuadros already approved of breaking changes in the API on a leads meeting 

Is there a chance we could link in relevant minutes where it has happened?
Could not find any with quick search of `breaking`.


> So if everything else works here and @vmarkovtsev is happy with Python API,

Did anybody hear back from Machine Learning team on these API changes ? Would appreciate if anyone can point me to the right place with the discussion.

A while ago has asked to be included in the conversation but did not get any feedback so far https://github.com/src-d/minutes/pull/372#discussion_r233136062

juanjux:
No idea what meeting was, one of the latest I was in, surely @mcuadros remembers it, I proposed to discuss if we had to support backward breaking changes and creating a compatibility layer and he said it wouldn't be needed since it was a major version increase. Maybe it wasn't recorded. 

We asked for ML feedback on the leads meeting and quoting several times here, but we had to go forward. Anyway the Python layer I added over the native one while not backwards compatible should be pretty easy to use.

juanjux:
@bzz everything in your review should be solved.

bzz:
@vmarkovtsev would you be so kind to please review the API changes?

setup.py
```diff
@@ -249,16 +297,27 @@ def main():
...
         extra_compile_args=["-std=c++11"],
-        include_dirs=[j("bblfsh", "libuast"), "/usr/local/include", "/usr/local/include/libxml2",
-                      "/usr/include", "/usr/include/libxml2"], sources=sources)
+        extra_objects=extra_objects,
+        include_dirs=[j("bblfsh", "libuast")],
+        sources=sources)
 
     setup(
-        cmdclass = {
+        cmdclass={
```
zurk: I believe pep8 want to see spaces here. :)

.travis.yml
```diff
@@ -1,22 +1,26 @@
...
 dist: xenial
+env: 
+- BBLFSHD_VERSION=v2.9.1 BBLFSH_PYTHON_VERSION=v2.3.0
 services:
   - docker
 cache:
   directories:
     - $HOME/.cache/pip
 python:
-  - "3.5"
```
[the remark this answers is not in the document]
juanjux: Only for Travis since this allows us to use full typing support in the tests.

bblfsh/client.py
```diff
@@ -60,54 +63,39 @@ def parse(self, filename: str, language: str=None, contents: str=None,
                          currently supported languages. None means autodetect.
         :param contents: The contents of the file. IF None, it is read from \
                          filename.
+        :param mode:     UAST transformation mode.
+        :param raw:      Return raw binary UAST without decoding it.
         :param timeout: The request timeout in seconds.
+        :type filename: str
```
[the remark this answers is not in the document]
juanjux: It's above.

bblfsh/result_context.py
```diff
@@ -0,0 +1,129 @@
...
+    def iterate(self, order: int) -> NodeIterator:
+        TreeOrder.check_order(order)
+        return NodeIterator(iterator(self._ctx.root(), order))
+
+    @property
+    def language(self) -> str:
+        return self._response.language
+
+    @property
+    def uast(self) -> t.Any:
```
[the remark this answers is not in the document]
juanjux: Because mypy had problems checking the type of some gRPC types as in this case.

bblfsh/fixtures/test.py
```diff
@@ -0,0 +1,322 @@
...
+        self.assertEqual(ctx.language, "python")
+
+    def testUASTFileContents(self):
+        with open(__file__, "rb") as fin:
+            contents = fin.read()
+        ctx = self.client.parse("file.py", contents=contents)
+        self._validate_ctx(ctx)
+        self._validate_filter(ctx)
+    #
+    # def testBrokenFilter(self):
```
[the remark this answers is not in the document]
juanjux: Yes, until an issue with libuast/SDK is fixed, see discussion above.

bblfsh/client.py
```diff
@@ -1,54 +1,57 @@
...
+
+        if isinstance(contents, bytes):
+            contents = BblfshClient._ensure_utf8(contents)
+
         return contents
 
-    def parse(self, filename: str, language: str=None, contents: str=None,
-              timeout: float=None) -> ParseResponse:
+    def parse(self, filename: str, language: t.Optional[str]=None,
+              contents: t.Optional[str]=None, mode: t.Optional[ModeType]=None,
```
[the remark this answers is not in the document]
juanjux: It's fine. A function taking a nullable union of str and bytes like `_get_contents` can definitely take a nullable str.

bblfsh/tree_order.py
```diff
@@ -0,0 +1,12 @@
+class TreeOrder:
```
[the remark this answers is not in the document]
juanjux: Or data classes, or named tuples, or... :) But will change it on the next commit to `IntEnum` to still allow int comparisons since gRPC will return raw ints.

setup.py
juanjux: I don't think so, removing the spaces gives on my Vim PEP8 checker:

`E251: Unexpected spaces around keyword / parameter equals`

zurk: oh, sure. sorry.

bblfsh/aliases.py
```diff
@@ -1,27 +1,31 @@
...
 # "in" is a reserved keyword in Python thus can't be used as package name, so
 # we import by string
-uast_module = importlib.import_module(
-    "bblfsh.gopkg.in.bblfsh.sdk.%s.uast.generated_pb2" % VERSION)
-protocol_module = importlib.import_module(
-    "bblfsh.gopkg.in.bblfsh.sdk.%s.protocol.generated_pb2" % VERSION)
-protocol_grpc_module = importlib.import_module(
-    "bblfsh.gopkg.in.bblfsh.sdk.%s.protocol.generated_pb2_grpc" % VERSION)
+uast_v2_module = importlib.import_module("bblfsh.gopkg.in.bblfsh.sdk.v2.uast.generated_pb2")
+protocol_v2_module = importlib.import_module("bblfsh.gopkg.in.bblfsh.sdk.v2.protocol.generated_pb2")
```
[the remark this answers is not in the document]
juanjux: Fixed.

bblfsh/client.py
```diff
@@ -60,54 +63,39 @@ def parse(self, filename: str, language: str=None, contents: str=None,
...
         return sup_response.languages
 
-    def version(self):
+    def version(self) -> VersionResponse:
         """
         Queries the Babelfish server for version and runtime information.
 
         :return: A dictionary with the keys "version" for the semantic version and
-                 "build" for the build timestamp.
+                 # "build" for the build timestamp.
```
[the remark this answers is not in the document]
juanjux: Yep, thanks.

README.md
```diff
@@ -49,21 +51,49 @@ Please, read the [getting started](https://doc.bblf.sh/using-babelfish/getting-s
...
+# are alterative typed versions:
+x = next(ctx.filter("boolean(//*[@strtOffset or @endOffset])").get_bool()
+y = next(ctx.filter("name(//*[1])")).get_str()
+z = next(ctx.filter("count(//*)").get_int() # or get_float()
+
+# You can also iterate using iteration orders different than the 
+# default preorder using the `iterate` method on `parse` result or node objects:
+
+# Directly over parse results
+iter = client.parse("/path/to/file.py").iterate(bblfsh.TreeOrder.POST_ORDER)
```
vmarkovtsev: `iter` is a reserved keyword, we should not use it for variable names

README.md
```diff
@@ -49,21 +51,49 @@ Please, read the [getting started](https://doc.bblf.sh/using-babelfish/getting-s
...
+    doSomething(node.get())
+
+# filter must be used when using XPath functions returning these types:
+# XPath queries can return different types (dicts, int, float, bool or str), 
+# calling get() with an item will return the right type, but if you must ensure
+# that you are getting the expected type (to avoid errors in the queries) there
+# are alterative typed versions:
+x = next(ctx.filter("boolean(//*[@strtOffset or @endOffset])").get_bool()
+y = next(ctx.filter("name(//*[1])")).get_str()
+z = next(ctx.filter("count(//*)").get_int() # or get_float()
```
vmarkovtsev: This is not very Pythonic. IDK if it is hard to guess the type, but the following is more Pythonic:

```python
x = next(ctx.filter("boolean(//*[@strtOffset or @endOffset])").iterate())
for name in ctx.filter("name(//*[1])").iterate():
    print(name)
```

Where `iterate()` would be an unordered iterator over the resulting values. Or name it `results()`. Or remove it and iterate directly, anyway.

README.md
```diff
@@ -49,21 +51,49 @@ Please, read the [getting started](https://doc.bblf.sh/using-babelfish/getting-s
...
+
+# Over individual node objects to change the iteration order of
+# a specific subtree:
+ctx = client.parse("file.py")
+first_node = next(ctx)
+newiter = first_node.iterate(bblfsh.TreeOrder.POSITION_ORDER)
+for i in newiter: ...
+
+# You can also get the non semantic UAST or native AST:
+ctx = client.parse("file.py", mode=bblfsh.ModeDict["NATIVE"])
```
vmarkovtsev: We should use an enum class or define separate constants instead of raw strings.

bblfsh/client.py
```diff
@@ -1,54 +1,57 @@
 import os
-import sys
+import typing as t
```
vmarkovtsev: Let's import individual types from `typing`. `t` is too short and error-prone while `typing` is too long for complex type defs.

bblfsh/client.py
```diff
@@ -1,54 +1,57 @@
...
 
         :param endpoint: The address of the Babelfish server, \
                          for example "0.0.0.0:9432"
+        :type endpoint: str
         """
+
         self._channel = grpc.insecure_channel(endpoint)
-        self._stub = ProtocolServiceStub(self._channel)
+        self._stub_v1 = ProtocolServiceStub(self._channel)
+        self._stub_v2 = DriverStub(self._channel)
```
vmarkovtsev: This is unrelated to the PR but maybe you can add at once:

```python
def close(self):
    """
    Close the gRPC channel and free the acquired resources. Using a closed client is not supported.
    """
    self._channel.close()
    self._channel = self._stub_v1 = self._stub_v2 = None
```

We need this function in many different places and currently call `_channel.close()`.

bblfsh/client.py
```diff
@@ -1,54 +1,57 @@
...
 
-class BblfshClient(object):
+class BblfshClient:
     """
-    Babelfish gRPC client. Currently it is only capable of fetching UASTs.
+    Babelfish gRPC client.
     """
 
-    def __init__(self, endpoint: str):
+    def __init__(self, endpoint:str) -> None:
```
vmarkovtsev: This is unrelated to the PR but here is another API request that we really need:

```python
@staticmethod
def wrap(channel, stub_v1, stub_v2) ->  "BblfshClient":
    client = BblfshClient.__new__()
    client._channel = channel
    client._stub_v1 = stub_v1
    client._stub_v2 = stub_v2

def __init__(self, endpoint: str):
    channel = grpc.insecure_channel(endpoint)
    stub_v1 = ProtocolServiceStub(channel)
    stub_v2 = DriverStub(channel)
    return BblfshClient.wrap(channel, stub_v1, stub_v2)
```

vmarkovtsev: (whitespace is needed before the type)
```suggestion
    def __init__(self, endpoint: str) -> None:
```

README.md
vmarkovtsev: We don't use those removed functions

vmarkovtsev:
So given that we no longer have a `Node`, we have to rewrite literally everything on our side, because `Node` is used everywhere in ML code, in nearly half of the files. Alternatively, we will write our own emulation wrapper because we heavily rely on that abstraction anyway and don't want to spend months on debugging very very tricky stuff bound to nodes :)

vmarkovtsev:
Daaaamn, we have to rewrite bloody months of work in the style-analyzer's feature extraction without this emulation.

juanjux:
@vmarkovtsev That was [my point of view](https://github.com/bblfsh/client-python/pull/128#discussion_r230415441), when I disliked the breaking changes. It doesn't make sense for you to write a `Node` compatibility layer because if there is any it should be here (`from bblfsh.past import Node` or something like that).

README.md
juanjux: Because you didn't use queries returning anything outside of nodes, but that doesn't mean they should be removed (for example the sonar checks used some of them IIRC).

README.md
juanjux: Those are typed functions for queries returning boolean/string/integer/float values instead of nodes so the "typed get" is needed when the user doesn't have to do `isinstance` of the results to check that they're right every time.

README.md
juanjux: It's mapping the `protocol_v2_module.Mode.DESCRIPTOR.values_by_name` field. Maybe we could generate variable names from these strings using `eval` or `exec` but that's almost worse.

bblfsh/client.py
juanjux: Done

bblfsh/client.py
juanjux: Can you explain what's the use case of this?

bblfsh/aliases.py
```diff
@@ -1,27 +1,36 @@
...
+
+DESCRIPTOR = uast_v2_module.DESCRIPTOR
+ParseRequest = protocol_v2_module.ParseRequest
+ParseResponse = protocol_v2_module.ParseResponse
+ParseError = protocol_v2_module.ParseError
+Mode = protocol_v2_module.Mode
+ModeType = google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper
+
+# Current values: {'DEFAULT_MODE': 0, 'NATIVE': 1, 'PREPROCESSED': 2, 'ANNOTATED': 4, 'SEMANTIC': 8}
+ModeDict = {}
```
vmarkovtsev:
```suggestion
class Modes:
    pass
for k, v in Mode.DESCRIPTOR.values_by_name.items():
    setattr(Modes, k, v)
```

juanjux: notbad.jpg :+1:

juanjux: (will commit separately since github doesn't allow me to commit suggestions to this branch)

bblfsh/client.py
juanjux: Done, thanks!

setup.py
creachadair: Drive-by: Do we (or if not then could we) use a formatter like [yapf](https://github.com/google/yapf), which support PEP8 style?

bzz: I belive ML team [might be using flake8](https://github.com/search?q=org%3Asrc-d+pep8&type=Issues)

creachadair: As far as I know, though, flake8 only _complains_ about errors, it doesn't fix them. For formatting questions, it'd be nice to have a tool that does the right thing.

dennwc:
Re "typed get" issue mentioned above, I agree with @vmarkovtsev - I don't see the reason for typed get helpers. Python is a dynamic language, so it should be okay to iterate over nodes directly (that already are strings/ints/whatever). Old client had those helpers because the way `libxml` was wrapped. It's not necessary right now.

juanjux:
Python is dynamically typed but not weakly typed so it's not considered good practice on Python to have variant return types from the same function other than None. Also, having typed functions make easier to use mypy and avoid doing "if isinstance(...)"  on every value, which would also be slower (or risking having buggy code using types without checking them). Plus is basically free to have them on the API.

dennwc:
@juanjux OK, thanks for clarifying it. I trust your decision here.

juanjux:
Update: I started to work on a compatiblity layer with v2 before the holidays, but it's stopped to work on the very prioritary C# driver, I'll continue with it ASAP once the driver is finished so we can finally publish this version.

juanjux:
Since this passed review from the team, I propose to merge it (WITHOUT releasing v3 yet) so the compatiblity layer PR will be much easier to review as a separate one. What do you think?

creachadair:
> Since this passed review from the team, I propose to merge it (WITHOUT releasing v3 yet) so the compatiblity layer PR will be much easier to review as a separate one. Both PRs, once merged, would then be released as V3. What do you think?

That sounds reasonable to me.

dennwc:
Sounds good. Let's merge it.

PR #129, opened 2018-10-09
fix broken links in README.md

vovak:
Links to UAST specification and the language clients section in the doc looked obsolete.


juanjux:
Thanks for the PR!

PR #132, opened 2018-10-23
Add upload to pypi from travis tags

juanjux:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,14 @@ install:
   - pip3 install -r requirements.txt
   - python3 setup.py --getdeps --log
   - pip3 install . --upgrade
+  - if [[ -z "$TRAVIS_TAG" ]]; then exit 0; fi
+  - echo "[distutils]" > $HOME/.pypirc
+  - echo "index-servers = " >> $HOME/.pypirc
+  - echo "  pypi" >> $HOME/.pypirc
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -17,6 +17,14 @@ install:
   - pip3 install -r requirements.txt
   - python3 setup.py --getdeps --log
   - pip3 install . --upgrade
+  - if [[ -z "$TRAVIS_TAG" ]]; then exit 0; fi
```
dennwc: Will it actually exit the script? For example, each Docker's `RUN` command executes in a separate script. Might not be the case with Travis, though.

.travis.yml
```diff
@@ -17,6 +17,14 @@ install:
...
+  - echo "[distutils]" > $HOME/.pypirc
```
dennwc: Maybe we can host the header of this file in the repo and add only username/password here?

.travis.yml
juanjux: Looks like it does (from this PR CI at least).

.travis.yml
juanjux: We could, but it's very small so this makes it one unrelated file less.

bzz:
Would be nice if this could read version from the "${TRAVIS_TAG}"..

PR #138, opened 2018-12-07
Use version range for grpcio instead of exact version

smacker:
Fixed version of grpcio is causing error when an application needs to use
newer compatible grpcio:
bblfsh 2.11.2 has requirement grpcio==1.13.0, but you'll have grpcio 1.17.0
which is incompatible.

In my case I stumbled on this bug that is fixed in newer grpcio:
https://github.com/grpc/grpc/issues/14088

Signed-off-by: Maxim Sukharev <max@smacker.ru>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -272,7 +272,7 @@ def main():
         packages=find_packages(),
         exclude=["bblfsh/test.py"],
         keywords=["babelfish", "uast"],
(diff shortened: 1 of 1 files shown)


juanjux:
We locked the version because of an issue, but I'm going to guess that will be already fixed.

PR #142, opened 2019-03-15
V1 compatibility layer

juanjux:
This PR adds a compatibility layer with the v1 API of this same clients (python-client v2 versions).

The layer uses the same objects and methods as the old API but the XPath expressions used would have to be updated since the projection is different.

It also fix some issues found on the existing code, both the the native CPython sources and in the pythonic layer over it.


juanjux:
Just noticed it crashes with Python 3.7 on Travis... will debug it.

bblfsh/compat.py
```diff
@@ -0,0 +1,228 @@
+import os
```
creachadair: Can you please add a comment to this file summarizing what it is and outlining the API?

Even if the answer is just "this is the v1 API surface (see link) and bridges to a v2 client", that would be helpful for the future reader.

bblfsh/compat.py
```diff
@@ -0,0 +1,228 @@
...
+from bblfsh.result_context import ResultContext
+from bblfsh.aliases import (
+    ParseRequest, ParseResponse, DriverStub, ProtocolServiceStub,
+    VersionRequest, SupportedLanguagesRequest, ModeType,
+    Mode, VersionResponse, DESCRIPTOR
+)
+from bblfsh.pyuast import uast, iterator as native_iterator
+from bblfsh.tree_order import TreeOrder
+
+if "BBLFSH_COMPAT_SHUTUP" not in os.environ:
```
creachadair: Haha, this name. 😁

Unless someone requested this, I'd be inclined to post the warning unconditionally. Otherwise this will get wired into somebody's setup script somewhere and never touched again. If someone comes requesting a silence for it, we can use that as an opportunity to work with them to get them migrated.

bblfsh/compat.py
```diff
@@ -0,0 +1,228 @@
...
+    print("Warning: using deprecated bblfsh v1 compatibility layer.",
+          file=sys.stderr)
+
+
+class WrongTypeException(Exception):
```
creachadair: Please add at least brief doc-strings to this and the other classes below. For exceptions I think the interesting datum is where the exception comes from (e.g., "this is thrown from queries when the API returns a type that doesn't match the request" or whatever is actually true).

bblfsh/compat.py
```diff
@@ -0,0 +1,228 @@
...
+    def ast(self) -> Node:
+        return self._res_context.ast
+
+    @property
+    def ctx(self) -> ResultContext:
+        return self._res_context
+
+    @property
+    def elapsed(self) -> int:
+        # FIXME(juanjux): check if the can get this or measure ourselves
```
creachadair:
```suggestion
        # FIXME(juanjux): check if the caller can get this, or measure it ourselves.
```

bblfsh/compat.py
```diff
@@ -0,0 +1,228 @@
...
+
+    def version(self) -> VersionResponse:
+        return self._bblfsh_cli.version()
+
+    def close(self) -> None:
+        return self._bblfsh_cli.close()
+
+
+class CompatNodeIterator:
+    def __init__(
```
creachadair: [nit] Even with an 80-column limit, this fits on one line. (Well, okay it's exactly 80 so the line break rolls over—calling the iterator `it` would fix that ;)

bblfsh/compat.py
```diff
@@ -0,0 +1,228 @@
...
+            self,
+            nodeit: NodeIterator,
+            only_nodes: bool = False
+    ) -> None:
+        self._nodeit = nodeit
+        self._only_nodes = only_nodes
+        # Used to forward calls of the old Node object
+        # Check if this, and properties(), are needed
```
creachadair: Is this meant to be a TODO?

bblfsh/compat.py
```diff
@@ -0,0 +1,228 @@
...
+        self._last_node: Optional[Node] = None
+
+    def __iter__(self) -> 'CompatNodeIterator':
```
creachadair: Why a string instead of the class itself? User-defined classes _should_ work as type annotations, but maybe there is something else going on in this case?

bblfsh/compat.py
```diff
@@ -0,0 +1,228 @@
...
+
+def iterator(n: Union[Node, CompatNodeIterator], order: TreeOrder = TreeOrder.PRE_ORDER)\
+        -> CompatNodeIterator:
+
+    if isinstance(n, CompatNodeIterator):
+        return CompatNodeIterator(n._nodeit.iterate(order), only_nodes=True)
+    elif isinstance(n, Node):
+        nat_it = native_iterator(n.internal_node, order)
+        return CompatNodeIterator(NodeIterator(nat_it), only_nodes=True)
+    elif isinstance(n, dict):
```
creachadair: Assuming the type annotations were actually checked, would this even be possible?

bblfsh/compat.py
juanjux: Will do.

bblfsh/compat.py
juanjux: Someone will probably request it... but let's remove it until then.

bblfsh/compat.py
juanjux: Sure.

bblfsh/compat.py
juanjux: Out [current conventions](https://github.com/src-d/guide/blob/master/engineering/conventions/python.md) specify a 99 line length so I'll merge all.

bblfsh/compat.py
juanjux: Yes; I've removed it (I already checked and they're needed).

bblfsh/compat.py
juanjux: You use a string with mypy annotations when it's a reference to the enclosing class (as is the case here) or other class that is not defined yet.

bblfsh/compat.py
juanjux: It's possible since the new API can work with normal dictionaries as nodes, but the `Union` parameter annotation was wrong since it was missing the dict; I'll fix it on the next commit.

bblfsh/compat.py
juanjux: Done!

bblfsh/compat.py
juanjux: Done!

juanjux:
@creachadair feedback handled, ready for another round.

bblfsh/node.py
```diff
@@ -44,7 +44,12 @@ def offset(self) -> int:
     def offset(self, v: int) -> None:
         self._parent_pos["offset"] = v
 
-
+# This is for v1 "node.children" compatibility. It will update the children
+# property with the dict or Node objects in properties or list/tuple properties
+# when .children is accessed (because the user could change the node using get_dict()
+# or .properties).
+# Also, all these " in children" are O(1) so this will be slow for frequently accessing
```
creachadair: Did you mean `O(n)` here?

bblfsh/compat.py
creachadair: Makes sense, thanks for the explanation!

bblfsh/node.py
juanjux: Yes, sorry, had a brainfart.

bzz:
Nice work 👏 

From the tests, I understood that on API level we expect downstream users to explicitly import `bblfsh.compat` and things from it, case they want to leverage this API - would that be correct?

And one more question, on the API namespace level that most probably is not directly related to this PR but rather to upcoming v3 as a whole.

WDYT, does it make sense to target drop-in replacement of the old API with a new version in the same `bblfsh` package or would it rather make sense to somehow allow for both be installed/imported/distinguished at the same time?

bblfsh/compat.py
```diff
@@ -0,0 +1,335 @@
...
+class CompatBblfshClient:
+    """
+    This emulates the methods and properties of the old BblfshClient.
+    """
+    def __init__(self, endpoint: Union[str, grpc.Channel]) -> None:
+        """
+        Connects to the specified grpc endpoint which can be specified either as
+        a grpc Channel object or a connection string (like "0.0.0.0:6432").
+        """
+        self._bblfsh_cli = bcli.BblfshClient(endpoint)
```
bzz: Probably a nitpick, but from the first glance the name `_bblfsh_cli ` made me think about the command line interface and why is it used everywhere.

juanjux: It's from cli-end, but now I see it could be confussing, I've renamed the alias to newbbl (it's only used internally anyway).

juanjux:
@bzz yes, the client would need to import from bblfsh.compat as the `test_compat.py` file does.

PR #151, opened 2019-04-19
Fix pypi upload

dennwc:
Try to fix the pypi upload issues.

Currently, it seems like username/password is not propagated correctly. So instead of writing it to `.pypirc` in the current dir, write it to `HOME` so it's accessible globally.

Also, change a few other things in CI config:
- Instead of hard-coding the version in repo, use `sed` in CI to replace the version with a current Git tag.
- Previously pipy upload was in `install` step, which is weird, since `install` is intended for dependencies. Move it to `after_success` step instead.


.travis.yml
```diff
@@ -16,22 +16,21 @@ install:
...
+after_success:
+  - if [[ -z "$TRAVIS_TAG" ]]; then exit 0; fi
+  - if [[ $TRAVIS_PYTHON_VERSION != '3.6' ]]; then exit 0; fi # disable double uploads to pypi
+  - echo "[distutils]" > $HOME/.pypirc
+  - echo "index-servers = " >> $HOME/.pypirc
+  - echo "  pypi" >> $HOME/.pypirc
+  - echo "[pypi]" >> $HOME/.pypirc
+  - echo "username=$PYPI_USER" >> $HOME/.pypirc
+  - echo "password=$PYPI_PASS" >> $HOME/.pypirc
+  - python setup.py sdist upload
```
creachadair:
```suggestion
  - python setup.py sdist upload
```

setup.py
```diff
@@ -12,7 +12,7 @@
 from setuptools import setup, find_packages, Extension
 from setuptools.command.build_ext import build_ext
 
-VERSION = "3.0.1"
+VERSION = "3.x.x-dev"
```
creachadair:
```suggestion
# The VERSION line is edited automatically during deployments.
# You may change the contents of the string, but do not otherwise edit the line.
VERSION = "3.x.x-dev"
```

.travis.yml
dennwc: Sorry, I don't see any change on this line.

setup.py
dennwc: Done

.travis.yml
dennwc: Assuming it's okay. Let me know if you want me to change something else.

creachadair: It was just adding the missing LF at EOF. Cosmetic, not crucial.

PR #154, opened 2019-04-24
Require Python ≥ 3.6.

### LINKED ISSUE #153: Syntax for variable annotations requires  python 3.6=>
https://github.com/bblfsh/client-python/blob/cc9fbc4cd52948cd23dea806c88af6381f0c84ce/bblfsh/node.py#L161

creachadair:
Some of our code uses type annotations on variable declarations, which were introduced by [PEP 526](https://www.python.org/dev/peps/pep-0526/) with Python 3.6. There does not seem to be any good reason not to prefer newer versions of Python 3, since we are already incompatible with Python 2, so this change drops the 3.4 and 3.5 tags from the PyPI configuration.

Fixes #153.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -336,8 +336,6 @@ def main():
             "Intended Audience :: Developers",
             "License :: OSI Approved :: Apache Software License",
             "Operating System :: POSIX",
-            "Programming Language :: Python :: 3.4",
-            "Programming Language :: Python :: 3.5",
             "Programming Language :: Python :: 3.6",
             "Programming Language :: Python :: 3.7",
             "Topic :: Software Development :: Libraries"


creachadair:
I guess in principle this may count as a breaking change. Should we also bump the version number?

dennwc:
A quick search only shows [this file](https://github.com/bblfsh/documentation/blob/master/using-babelfish/clients.md), but it doesn't mention the version.

Regarding the breaking change, I think we already bumped the major version when switching to SDKv3, so it should be fine.

creachadair:
> I think we already bumped the major version when switching to SDKv3, so it should be fine.

Ok, thanks. In that case I will merge this once I have updated the description.

PR #172, opened 2019-07-23
Fix classes for backward compatibility

zurk:
Hey, LA team 👋 

ML team found some time to check out new client (v3) you provide for us. Here a PR to make everything work as it was. https://github.com/src-d/ml-core/pull/31 is the PR in `ml-core` that is waiting for a new release of bblfsh client. 

It is more or less obvious what I changed and why, but ask me while reviewing if it is unclear. Every change is necessary or some test fails in ml-core`.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/172)

### DIFF
--- a/bblfsh/__init__.py
+++ b/bblfsh/__init__.py
@@ -2,25 +2,4 @@ from bblfsh.client import BblfshClient
 from bblfsh.pyuast import decode, iterator, uast
 from bblfsh.tree_order import TreeOrder
 from bblfsh.aliases import *
-
-
-class RoleSearchException(Exception):
-    pass
-
-
-def role_id(rname: str) -> int:
-    try:
-        name = DESCRIPTOR.enum_types_by_name["Role"].values_by_name[rname].number
-    except KeyError:
-        raise RoleSearchException("Role with name '{}' not found".format(rname))
-
-    return name
-
-
-def role_name(rid: int) -> str:
-    try:
-        id_ = DESCRIPTOR.enum_types_by_name["Role"].values_by_number[rid].name
-    except KeyError:
-        raise RoleSearchException("Role with ID '{}' not found".format(rid))
-
-    return id_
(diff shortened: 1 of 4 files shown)


zurk:
I cannot assign reviewers, so
cc @creachadair

zurk:
yes, it was yesterday too. I just started to fix them, but I need your help because there is C extension involved and I am not familiar with it.

https://travis-ci.com/bblfsh/python-client/jobs/218999263#L332

iterator test fails. But if you change a position of the root node to 1 (`root.start_position.offset = 1`)
It works! 

Any ideas what is wrong are welcome.

zurk:
also If you take current master and change default positions from -1 to 0 (as it should be):
https://github.com/bblfsh/python-client/blob/f2a7d7ee9b0ffa6ae24fbad1af2013d48687a0c6/bblfsh/node.py#L235-L248

test also fails.

zurk:
ok, I think this bug is unrelated to my changes and located somewhere in C extension code.
So, I fix the test to make it pass as it was. 
If you want I can create an issue about the iterator I found.

zurk:
that is true, but when we have a protobuf and the node did not have positions all three values were zeros. I agree that it is not the best way to put things, but some of our code rely on (0,0,0) position.

zurk:
Ah! So you want to say that this `testIteratorPositionOrder` puts Nodes with invalid positions to the end? In this case, how iterator knows which position is invalid?

zurk:
Ok, waiting for a merge (if everything is fine) and a new release so I can use all these changes in ml-core. 🤞

zurk:
@creachadair friendly ping :)

zurk:
Thanks! @dennwc Are we good to merge?
I am not familiar with a tool you use for a review but it is still pending

PR #176, opened 2019-07-29
Fixes Python object caching memory leaks

ncordon:
* ~~Deletes unbalanced `Py_INCREF` [here](https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/pyuast.cc#L347), and analogously for `PythonContext_filter`. The rationale behind this is that this `Py_INCREF` has no effect because of `NodeIterator` [saving a pointer to `ctx`](https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/node_iterator.py#L19). Another alternative would be to do a `Py_DECREF` in `PyUastIterExt_dealloc`, but `PyUastIterExt` is also used for free contexts which are not associated to a Python object, so I think this is the cleanest solution.~~ This would be addressed separately #181,
* Python shares ints and strings in memory (for ints I think it is in a certain range [0,256]). Let us suppose we try to create two Python objects which hold the integer `26` (-> we would be returned the same reference, but with count 2). Thus, when we create a new node associated to the two integers [here](https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/pyuast.cc#L627), we are only storing a single node (we are overwriting it) , so we would only decrease `26` one time when we free `Interface`. I now store a list of the creates nodes to free them when `Interface` is released, and I cache only the last one corresponding to a certain Python object. This would work as long as we are not planing to let the user modify a node in the tree (?).

Needs https://github.com/bblfsh/libuast/pull/107. The former, #181 and this PR solve #176 

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/176)

### DIFF
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -824,9 +824,12 @@ public:
         ctx = impl->NewContext();
     }
     ~Context(){
+        // Deleting ctx also frees impl
         delete(ctx);
-        delete(impl);
+        ctx = nullptr;
+        impl = nullptr;
         delete(iface);
+        iface = nullptr;
     }
 
     // RootNode returns a root UAST node, if set.
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -601,10 +601,10 @@ class Context;
 class Interface : public uast::NodeCreator<Node*> {
 private:
     // Thread safety is ensured here by Python's GIL (Global Interpreter Lock), since only
-    // a method could enter in a C++ routine at a time. Note that the STL container used
-    // to cache the objects does not ensure two concurrent writes thread safety, but GIL in
-    // this case indirectly provides us that feature. Also ensures thread safety for the
-    // lookupOrCreate and createIfNotExists methods
+    // a method could enter in a C++ routine at a time.
+    // The STL container used to cache the objects does not support concurrent writes, but
+    // Python's GIL (Global Interpreter Lock) ensures lookupOrCreate and createIfNotExists
+    // will not execute concurrently, avoiding concurrent writers to the cache.
     std::unordered_map<PyObject*, Node*> obj2node;
 
     static PyObject* newBool(bool v) {
@@ -676,14 +676,14 @@ public:
 
     Node* NewObject(size_t size) {
         // Object is going to be created and cached, since when
-        // we call PyDict_New we are retruned a new reference
+        // we call PyDict_New we returned a new reference
         // and therefore we are not going to use a cached one.
         PyObject* m = PyDict_New();
         return createIfNotExists(NODE_OBJECT, m);
     }
     Node* NewArray(size_t size) {
         // Array is going to be created and cached, since when
-        // we call PyList_New we are retruned a new reference
+        // we call PyList_New we returned a new reference
         // and therefore we are not going to use a cached one.
         PyObject* arr = PyList_New(size);
         return createIfNotExists(NODE_ARRAY, arr);
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -1,7 +1,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
-#include <map>
+#include <unordered_map>
 
 #include <Python.h>
 #include <structmember.h>
@@ -600,7 +600,7 @@ class Context;
 
 class Interface : public uast::NodeCreator<Node*> {
 private:
-    std::map<PyObject*, Node*> obj2node;
+    std::unordered_map<PyObject*, Node*> obj2node;
 
     static PyObject* newBool(bool v) {
         if (v) Py_RETURN_TRUE;
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -345,6 +345,7 @@ static PyObject *PythonContextExt_filter(PythonContextExt *self, PyObject *args,
         PyErr_SetString(PyExc_RuntimeError, e.what());
     }
 
+    Py_INCREF((PyObject *)self);
     return it;
 }
 
@@ -933,6 +934,7 @@ static PyObject *PythonContext_filter(PythonContext *self, PyObject *args, PyObj
         PyErr_SetString(PyExc_RuntimeError, e.what());
     }
 
+    Py_INCREF((PyObject *)self);
     return it;
 }
 
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -600,6 +600,11 @@ class Context;
 
 class Interface : public uast::NodeCreator<Node*> {
 private:
+    // Thread safety is ensured here by Python's GIL (Global Interpreter Lock), since only
+    // a method could enter in a C++ routine at a time. Note that the STL container used
+    // to cache the objects does not ensure two concurrent writes thread safety, but GIL in
+    // this case indirectly provides us that feature. Also ensures thread safety for the
+    // lookupOrCreate and createIfNotExists methods
     std::unordered_map<PyObject*, Node*> obj2node;
 
     static PyObject* newBool(bool v) {
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -631,6 +631,9 @@ private:
         Node* node = obj2node[obj];
         // This object's node is already cached; release the reference.
         if (node) {
+            // It is safe to DECREF here, since an INCREF has already happened at
+            // NewString (in PyUnicode_FromString), NewInt (in PyLong_FromLongLong),
+            // NewUint, NewFloat or NewBool
             Py_DECREF(obj);
             return node;
         }
@@ -667,10 +670,16 @@ public:
     }
 
     Node* NewObject(size_t size) {
+        // Object is going to be created and cached, since when
+        // we call PyDict_New we are retruned a new reference
+        // and therefore we are not going to use a cached one.
         PyObject* m = PyDict_New();
         return createIfNotExists(NODE_OBJECT, m);
     }
     Node* NewArray(size_t size) {
+        // Array is going to be created and cached, since when
+        // we call PyList_New we are retruned a new reference
+        // and therefore we are not going to use a cached one.
         PyObject* arr = PyList_New(size);
         return createIfNotExists(NODE_ARRAY, arr);
     }
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -2,7 +2,6 @@
 #include <cstdlib>
 #include <cstring>
 #include <unordered_map>
-#include <list>
 
 #include <Python.h>
 #include <structmember.h>
@@ -601,7 +600,6 @@ class Context;
 class Interface : public uast::NodeCreator<Node*> {
 private:
     std::unordered_map<PyObject*, Node*> obj2node;
-    std::list<Node*> nodes;
 
     static PyObject* newBool(bool v) {
         if (v) Py_RETURN_TRUE;
@@ -622,12 +620,22 @@ private:
         return node;
     }
 
-    // create makes a new object with a specified kind.
-    // Steals the reference.
-    Node* create(NodeKind kind, PyObject* obj) {
-        Node* node = new Node(this, kind, obj);
+    // create makes a new object with a specified kind, as long as it does
+    // not already exists. If it does, it just outputs the existing Node
+    // and decreases the ref count to obj
+    // Steals the reference iff the object does not already exists
+    Node* createIfNotExists(NodeKind kind, PyObject* obj) {
+        if (!obj || obj == Py_None) return nullptr;
+
+        Node* node = obj2node[obj];
+        // If obj is already cached, we do not want another reference to it
+        if (node) {
+            Py_DECREF(obj);
+            return node;
+        }
+
+        node = new Node(this, kind, obj);
         obj2node[obj] = node;
-        nodes.push_back(node);
         return node;
     }
 public:
@@ -639,8 +647,8 @@ public:
     ~Interface(){
         // Only needs to deallocate Nodes, since they own
         // the same object as used in the map key.
-        for (auto node : nodes)
-            delete(node);
+        for (auto it : obj2node)
+            delete(it.second);
     }
 
     // toNode creates a new or returns an existing node associated with Python object.
@@ -659,31 +667,31 @@ public:
 
     Node* NewObject(size_t size) {
         PyObject* m = PyDict_New();
-        return create(NODE_OBJECT, m);
+        return createIfNotExists(NODE_OBJECT, m);
     }
     Node* NewArray(size_t size) {
         PyObject* arr = PyList_New(size);
-        return create(NODE_ARRAY, arr);
+        return createIfNotExists(NODE_ARRAY, arr);
     }
     Node* NewString(std::string v) {
         PyObject* obj = PyUnicode_FromString(v.data());
-        return create(NODE_STRING, obj);
+        return createIfNotExists(NODE_STRING, obj);
     }
     Node* NewInt(int64_t v) {
         PyObject* obj = PyLong_FromLongLong(v);
-        return create(NODE_INT, obj);
+        return createIfNotExists(NODE_INT, obj);
     }
     Node* NewUint(uint64_t v) {
         PyObject* obj = PyLong_FromUnsignedLongLong(v);
-        return create(NODE_UINT, obj);
+        return createIfNotExists(NODE_UINT, obj);
     }
     Node* NewFloat(double v) {
         PyObject* obj = PyFloat_FromDouble(v);
-        return create(NODE_FLOAT, obj);
+        return createIfNotExists(NODE_FLOAT, obj);
     }
     Node* NewBool(bool v) {
         PyObject* obj = newBool(v);
-        return create(NODE_BOOL, obj);
+        return createIfNotExists(NODE_BOOL, obj);
     }
 };
 
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -833,13 +833,9 @@ public:
         ctx = impl->NewContext();
     }
     ~Context(){
-        // impl gets deleted when deleting ctx
-        delete(ctx);
-        delete(impl);
-        delete(iface);
-        ctx = nullptr;
-        impl = nullptr;
-        iface = nullptr;
+        delete(ctx); ctx = nullptr;
+        delete(impl); impl = nullptr;
+        delete(iface); iface = nullptr;
     }
 
     // RootNode returns a root UAST node, if set.
--- a/bblfsh/pyuast.cc
+++ b/bblfsh/pyuast.cc
@@ -2,6 +2,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <unordered_map>
+#include <list>
 
 #include <Python.h>
 #include <structmember.h>
@@ -344,7 +345,7 @@ static PyObject *PythonContextExt_filter(PythonContextExt *self, PyObject *args,
     } catch (const std::exception& e) {
         PyErr_SetString(PyExc_RuntimeError, e.what());
     }
-    Py_INCREF((PyObject *)self);
+
     return it;
 }
 
@@ -494,7 +495,6 @@ public:
         }
         if (obj) Py_DECREF(obj);
         if (str) delete str;
-
     }
 
     PyObject* toPy();
@@ -508,8 +508,7 @@ public:
             str = new std::string(s);
         }
 
-        std::string* s = new std::string(*str);
-        return s;
+        return str;
     }
     int64_t AsInt() {
         long long v = PyLong_AsLongLong(obj);
@@ -601,6 +600,7 @@ class Context;
 class Interface : public uast::NodeCreator<Node*> {
 private:
     std::unordered_map<PyObject*, Node*> obj2node;
+    std::list<Node*> nodes;
 
     static PyObject* newBool(bool v) {
         if (v) Py_RETURN_TRUE;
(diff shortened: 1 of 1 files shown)


ncordon:
___
*[bblfsh/pyuast.cc, line 641 at r1](https://reviewable.io/reviews/bblfsh/python-client/176#-LkxKhgg26yfrX0qfIq-:-LkxgU4T43BFcMzuaExn:b-ntyxgj) ([raw file](https://github.com/bblfsh/python-client/blob/31692ea9df80454e4658e5fedc433c0f659d5823/bblfsh/pyuast.cc#L641)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

Why do we need another container just to keep nodes for further delete?
Is ordering important here?
</blockquote></details>

Ordering is not important :). Do you suggest other way of keeping track of the nodes?

ncordon:
It was my fault that tests did not pass. Now they should (with version v3.3.1 of libuast).

However for the new version of libuast they may fail (master python-client +  master libuast also fails in the same test cases, so it seems it wasn't me).  We should look into that, because either we introduced a bug in libuast from v3.3.1 onwards (and there isn't any test that caught it there) or we have outdated tests here in python-client.

ncordon:
___
*[bblfsh/pyuast.cc, line 641 at r1](https://reviewable.io/reviews/bblfsh/python-client/176#-LkxKhgg26yfrX0qfIq-:-LkyOCOB7_KMT-M5tQed:bs3xwxl) ([raw file](https://github.com/bblfsh/python-client/blob/31692ea9df80454e4658e5fedc433c0f659d5823/bblfsh/pyuast.cc#L641)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

I've just thought that iterating through the map, deleting nodes , and erase the map, may do the job. Am I missing sth. here?
</blockquote></details>

Problem is that you may have lost nodes along the way if you just use the map. Let me give an example. Python caches integers from [0,256] internally. I parse the tree and I am going to find two 2s along the way. I do obj = newPyInt(2) and then obj2Node[obj] = new Node(obj, ...). Then again, in the same tree, obj = newPyInt(2) (a reference to the same object as before, but with the count increased). Then obj2Node[obj] = new Node(obj, ...) (I have lost the first Node turning into a subtle memory leak). Solution for me seems to either:
* Store all nodes in the list and release them all.
* Call lookupOrCreate instead of create (maybe this one is clearer).

What do you think?

ncordon:
___
*[bblfsh/pyuast.cc, line 641 at r1](https://reviewable.io/reviews/bblfsh/python-client/176#-LkxKhgg26yfrX0qfIq-:-LkyUmRO8gK86t9eoIBT:bs7wub4) ([raw file](https://github.com/bblfsh/python-client/blob/31692ea9df80454e4658e5fedc433c0f659d5823/bblfsh/pyuast.cc#L641)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

 I belive the whole idea of having a map was to create Nodes only once and then use it as a cache though `lookupOrCreate` so there should be no lost Nodes.
</blockquote></details>

Done!

ncordon:
___
*[bblfsh/pyuast.cc, line 623 at r4](https://reviewable.io/reviews/bblfsh/python-client/176#-Lkyp-3J3-9ISIhjzvlD:-LkysojWEgH82iwQcDd2:blw70e8) ([raw file](https://github.com/bblfsh/python-client/blob/0206795a954278b86dfdd529d8dd2e0d7a0cec2e/bblfsh/pyuast.cc#L623)):*

I think that maybe this comment isn't completely correct. Specifically, I think we're not actually _reducing_ the refcount of the existing object, we're just not _incrementing_ it (which ordinarily we would, if we picked it up naïvely from the cache).

If that's correct, then I suggest the following:

```
// createIfNotExists creates a new object with the specified kind if it does not
// already exist, and adds it to the cache. If the object does exist, it borrows 
// the existing reference from the cache without incrementing its reference count.
```

or words to that effect. Or if I'm wrong, please correct me.
</blockquote></details>

If we are parsing the tree and find two nodes associated with the number `x` in the UAST, we are going to try and do:

```
obj = newPythonInt(x);
createNode(obj); // then we create node = new Node(obj) since it does not already exists. And we store the relation obj -> node

// Again we have to create a node for x
. . . 
obj = newPythonInt(x); //it does not create another object. It just increases the ref count of the former one and returns me the same pointer
createNode(obj); 
// Two possibilities here, since obj already exists in the list. Either I output the same node as before, which is what we are                     
// doing right now, and we have to decrease the ref count of obj (since node is going to be released only once and its underlying
// PyObject DECREFed only once. Thi is safe to do, because all nodes are released at once, when Interface is destroyed). The other 
// possibility is to actually create another  node, and then I have to store all the nodes in a list, because I can have two or more nodes 
// associated to the same Python object and a map obj -> Node(obj) is just not enough to release all that memory. 
```

ncordon:
___
*[bblfsh/pyuast.cc, line 633 at r4](https://reviewable.io/reviews/bblfsh/python-client/176#-LkypgYG5pYRiv6tUfOF:-Lkyw9KR2CF42DbbC2s4:b-ltcmix) ([raw file](https://github.com/bblfsh/python-client/blob/0206795a954278b86dfdd529d8dd2e0d7a0cec2e/bblfsh/pyuast.cc#L633)):*

Why do we need to explicitly DECREF here? Does the `obj2node` lookup implicitly INCREF? (Operator overloading hides all the fun parts)
</blockquote></details>

I hope my last comment answers you also this  :smile_cat:

ncordon:
___
*[bblfsh/pyuast.cc, line 837 at r4](https://reviewable.io/reviews/bblfsh/python-client/176#-LkyqlTbAsOiXDvU0TMY:-LkyxgnND92kuaUiRmbs:b5sgb5r) ([raw file](https://github.com/bblfsh/python-client/blob/0206795a954278b86dfdd529d8dd2e0d7a0cec2e/bblfsh/pyuast.cc#L837)):*
> delete(ctx); ctx = nullptr;
> delete(impl); impl = nullptr;
> delete(iface); iface = nullptr;

Done! Probably @dennwc had it like that. But I have done so many experiments trying to find where the code leaked memory that I turned this part uglier.

ncordon:
___
*[bblfsh/pyuast.cc, line 623 at r4](https://reviewable.io/reviews/bblfsh/python-client/176#-Lkyp-3J3-9ISIhjzvlD:-LkyyzjJ7csqNIBxnTXN:b1tvlbd) ([raw file](https://github.com/bblfsh/python-client/blob/0206795a954278b86dfdd529d8dd2e0d7a0cec2e/bblfsh/pyuast.cc#L623)):*

If we are parsing the tree and find two nodes associated with the number `x` in the UAST, we are going to try and do:

```
obj = newPythonInt(x);
createNode(obj); // then we create node = new Node(obj) since it does not already exists. And we store the relation obj -> node

// Again we have to create a node for x
. . . 
obj = newPythonInt(x); //it does not create another object. It just increases the ref count of the former one and returns me the same pointer
createNode(obj); 
// Two possibilities here, since obj already exists in the list. Either I output the same node as before, which is what we are                     
// doing right now, and we have to decrease the ref count of obj (since node is going to be released only once and its underlying
// PyObject DECREFed only once. Thi is safe to do, because all nodes are released at once, when Interface is destroyed). The other 
// possibility is to actually create another  node, and then I have to store all the nodes in a list, because I can have two or more nodes 
// associated to the same Python object and a map obj -> Node(obj) is just not enough to release all that memory. 
```

</blockquote></details>

I forgot to say I tested it without `DECREF`ing the object, and it was losing memory

ncordon:
___
*[bblfsh/pyuast.cc, line 623 at r4](https://reviewable.io/reviews/bblfsh/python-client/176#-Lkyp-3J3-9ISIhjzvlD:-Ll1K7OyBZ6ffgGZ0VcH:b9y5yuh) ([raw file](https://github.com/bblfsh/python-client/blob/0206795a954278b86dfdd529d8dd2e0d7a0cec2e/bblfsh/pyuast.cc#L623)):*

Thank you for the detailed comment. I'm still not sure I get it, though, sorry.

There are two "interesting" paths through this function:

1. `obj2node[obj]` yields a pointer, indicating we already have constructed the object. In this case we are borrowing the reference from the cache: So the INCREF that was done when the object was originally created should still be the only one. That is: Unless `obj2node[obj]` hides another INCREF in the overload?

2. `obj2jnode[obj]` yields null, indicating this object does not exist yet. In this case we fall through to construct a new one. I assume the `Node` constructor must contain an INCREF somewhere, but maybe that's wrong.

From your explanation, it sounds like the cache is doing some reference manipulation. If that's correct, we should probably mention that in a comment here, e.g., 

```
// The cache lookup implicitly calls INCREF, so we have to DECREF to borrow
// rather than retaining another reference.
```

And more generally, it worries me that the reference-counting logic is spread out in different locations where we can't easily see the association. It might not be practical to make this change now, but ideally the two should be lexically "nearby" each other (e.g., create the node, INCREF, put it in the cache, done) rather than implicit (e.g., create the node, put it in the cache which implicitly INCREFs, then magically DECREF here and also during the delete operator for the cache, or whatever).

I know you tested it, so I didn't mean I think the code is incorrect; but rather that its correctness isn't obvious. :smile:


</blockquote></details>

Sorry for my incomplete explanations. I think sometimes when something starts making sense to me I suddenly stop making sense to everyone else :laughing: . From your reasoning, which is correct, there are two paths:
1. we have seen the object already, and there is no need to do `INCREF`
2. we have not seen the object already, and we need to do the `INCREF`.

That is correct, the only missing piece is that the `createIfNotExists` method receives the object with an `INCREF` already done (no matter if we had seen the object before or not). `createIfNotExists` is only called from the `New{Int, Float, Object,...}` methods (e.g. [an example](https://github.com/ncordon/python-client/blob/3c440aeb9992976ea1ae02337b7216274ee339b7/bblfsh/pyuast.cc#L681)). Those methods perform a`PyObject* obj = PyLong_FromLongLong // or PyFloat_FromDouble, ...`. Those calls to `Py{Whatever}` from the Python API already do an `INCREF` internally. So to compensate for that `INCREF`, we have to do a `DECREF` in the first case, and do nothing in the second one.

Better? If I am still not making sense, please tell me so. Also, I may of course be wrong about this reasoning.

ncordon:
___
*[bblfsh/pyuast.cc, line 637 at r5](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll1e0ocAiO3MSGPLjII:-Ll2Q_10AzZMFetx5zjP:bmkya0w) ([raw file](https://github.com/bblfsh/python-client/blob/3c440aeb9992976ea1ae02337b7216274ee339b7/bblfsh/pyuast.cc#L637)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

Can somebody please help me understand how thread-safe is this approach now? 

It might not be the case for Python due to GIL, but will be the case for JVM - what is going to happen if this method is called from 2 different native threads?

\cc @dennwc as am not sure whether this was a design consideration

</blockquote></details>

What do you mean by 2 different native threads? Does JVM cache its own objects as Python does? Can you please elaborate more on this question @bzz?

ncordon:
___
*[bblfsh/pyuast.cc, line 623 at r4](https://reviewable.io/reviews/bblfsh/python-client/176#-Lkyp-3J3-9ISIhjzvlD:-Ll6Nj68BXAntC8krp1p:b5yyjt7) ([raw file](https://github.com/bblfsh/python-client/blob/0206795a954278b86dfdd529d8dd2e0d7a0cec2e/bblfsh/pyuast.cc#L623)):*

Ah! I see now—we are not really _creating_ `obj`, only recording its existence, so it already has a positive refcount by the time this function is called. Thank you for your patience walking me through it.

So in summary: If we find it in the cache, we know that the refcount is already n ≥ 2 (n ≥ 1 for the object in the cache, plus the 1 added by the caller of this function) so we're just discarding the caller's reference. And since we know n ≥ 2 we can be sure our DECREF here will not invalidate the object.

If all of that is (finally) right, then I have a suggestion for how to document this function, which you should feel free to edit as you like:

```c
// createIfNotExists creates and caches a node corresponding to the given object, if 
// one does not already exist. This function takes ownership of the caller's reference
// to obj: If a new node is created, the reference is delegated to the cache; otherwise
// it is released.
```

And then:

```c
// This object's node is already cached; release the reference.
```

or words to that effect.

</blockquote></details>

@dennwc because the `INCREF`s are done once I lift the node to Python: https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/pyuast.cc#L653 Also, we cannot avoid the first `INCREF`: we do not have another way of seeing if something is in Python cache until you try and create the object and Python outputs you some reference it already has in its internal cache (with the `INCREF` already done).

ncordon:
___
*[bblfsh/pyuast.cc, line 623 at r4](https://reviewable.io/reviews/bblfsh/python-client/176#-Lkyp-3J3-9ISIhjzvlD:-Ll6Q-3X80UTUpUAi52p:b-48boqo) ([raw file](https://github.com/bblfsh/python-client/blob/0206795a954278b86dfdd529d8dd2e0d7a0cec2e/bblfsh/pyuast.cc#L623)):*

@dennwc because the `INCREF`s are done once I lift the node to Python: https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/pyuast.cc#L653 Also, we cannot avoid the first `INCREF`: we do not have another way of seeing if something is in Python cache until you try and create the object and Python outputs you some reference it already has in its internal cache (with the `INCREF` already done).
</blockquote></details>

@creachadair I have improved the comments of the method with your suggestions :)

ncordon:
___
*[bblfsh/pyuast.cc, line 670 at r5](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll2aJTb7WEidVmjl3m3:-Ll6Q_ZV7hBccW6l0IEC:b7ibys7) ([raw file](https://github.com/bblfsh/python-client/blob/3c440aeb9992976ea1ae02337b7216274ee339b7/bblfsh/pyuast.cc#L670)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

@dennwc whatever I do, I can not find this comments in reviewable.

Basically this is 1/2 of the fix in this PR, and there is the prior discussion https://github.com/bblfsh/python-client/pull/176#issuecomment-516048737

Otherwise Nacho proved it to be possible to loose pointers to some nodes (overwritten in cache) and leak the memory.
</blockquote></details>

Yes. I think the best possible way to explain this is: since Python is reusing the same object (for some integers, strings, etc it already has in memory), we should do the same with the C++ nodes associated to those objects. Otherwise we end up with two or more nodes which contain the same Python object and we have to keep track of those nodes and cannot have a bijection `obj -> Node(obj)` in our cache.

ncordon:
___
*[bblfsh/pyuast.cc, line 670 at r5](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll2aJTb7WEidVmjl3m3:-Ll6wHJKB7l_NFAs8kyB:bsugert) ([raw file](https://github.com/bblfsh/python-client/blob/3c440aeb9992976ea1ae02337b7216274ee339b7/bblfsh/pyuast.cc#L670)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

> I'm talking specifically about objects and arrays. Those cannot and shouldn't be reused.

Sorry, it was completely not clear to me! Could you please elaborate why?
</blockquote></details>

@dennwc Yeah, I completely agree with that, but they are not going to be reused anyway, since the reference we are going to get in its creation is not going to be the same. If you are suggesting splitting the implementation for creating arrays and objects and for everything else, I am also good with that.

ncordon:
We decided to fix the `Py_INCREF`s of the filter methods in a separate PR (see issue #181) and clarify in a comment that objects and arrays are not going to be reused at `Node` creation.

ncordon:
___
*[bblfsh/pyuast.cc, line 926 at r5](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll6tj0m9_D_-2c12ChW:-Ll7Fwx2DaD1SvgksQ0V:b-3zj0pg) ([raw file](https://github.com/bblfsh/python-client/blob/f2a7d7ee9b0ffa6ae24fbad1af2013d48687a0c6/bblfsh/pyuast.cc#L926)):*

Are you sure this is safe?

The reason we do this is because `Context` is now stored in the iterator. Imagine this:

```
it = client.parse("file").filter("//")
```

`ctx` here is not referenced by Python, thus it will try to destroy the reference. But it is not correct - iterator still has a reference to `ctx` in C, which Python cannot see. This why we should `INCREF` here and `DECREF` in the iterator's destructor.
</blockquote></details>

I have turned this into another issue #181 and rolled back the `INCREF`s

ncordon:
___
*[bblfsh/pyuast.cc, line 634 at r7](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll7e3GpDhQ8DFaAj1B4:-Ll7qTNHB42NZWcjwLIk:b-896fja) ([raw file](https://github.com/bblfsh/python-client/blob/1cfe17c96ffd7b68b8e7a246fb6f1f8e392a9239/bblfsh/pyuast.cc#L634)):*

Can you please also add some comment here about a specific place that does `INCREF` that corresponds to this `DECREF`?
</blockquote></details>

Done!

ncordon:
___
*[bblfsh/pyuast.cc, line 637 at r5](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll1e0ocAiO3MSGPLjII:-Ll7rCKNDxBgvLVrctc5:b-geigdt) ([raw file](https://github.com/bblfsh/python-client/blob/3c440aeb9992976ea1ae02337b7216274ee339b7/bblfsh/pyuast.cc#L637)):*

We also decided to add a comment somewhere to explain the expectations in regards to GIL/concurrency.
</blockquote></details>

Is this good enough in methods that create the nodes?:

```
    // For two threads working each one in its own context, this method is thread-safe, but
    // we cannot ensure thread-safety for two threads working concurrently in the same context
```

ncordon:
___
*[bblfsh/pyuast.cc, line 670 at r5](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll2aJTb7WEidVmjl3m3:-Ll7rNnD5QcmyHED3rM7:b5hmogc) ([raw file](https://github.com/bblfsh/python-client/blob/3c440aeb9992976ea1ae02337b7216274ee339b7/bblfsh/pyuast.cc#L670)):*

Apart from the comments in this PR, or in the next one?
</blockquote></details>

Well, I wrote this before our call, so I think with the comments it should suffice for a good  ease of maintenance

ncordon:
___
*[bblfsh/pyuast.cc, line 637 at r5](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll1e0ocAiO3MSGPLjII:-Ll7zGsq9QZEmotmIx3Y:b-mmw64w) ([raw file](https://github.com/bblfsh/python-client/blob/3c440aeb9992976ea1ae02337b7216274ee339b7/bblfsh/pyuast.cc#L637)):*

Mostly. Also, worth noting that the `obj2node` map is not thread-safe, but we rely on the fact that Python has a Global Interpreter Lock, so everything will be running in a single-thread mode.
</blockquote></details>

I have been reading about GIL, and have been playing with the bytecode of Python for 10 minutes. I think GIL ensures even thread safety here (for two threads working with the same context), since only one can call its `built-in` filter method (or `iterate` or whatever) and only one can make writes / reads from the cache (they cannot interleave). So my comment was not accurate enough

ncordon:
___
*[bblfsh/pyuast.cc, line 637 at r5](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll1e0ocAiO3MSGPLjII:-Ll8-prQ8KfWne9qQ4R6:b-gk1pca) ([raw file](https://github.com/bblfsh/python-client/blob/3c440aeb9992976ea1ae02337b7216274ee339b7/bblfsh/pyuast.cc#L637)):*

I have been reading about GIL, and have been playing with the bytecode of Python for 10 minutes. I think GIL ensures even thread safety here (for two threads working with the same context), since only one can call its `built-in` filter method (or `iterate` or whatever) and only one can make writes / reads from the cache (they cannot interleave). So my comment was not accurate enough
</blockquote></details>

Added the comment!

ncordon:
___
*[bblfsh/pyuast.cc, line 679 at r9](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll88x70-Uxdn-2HyE6T:-LlBldMz6Vg5TdZGwCMD:bs36yvc) ([raw file](https://github.com/bblfsh/python-client/blob/23c22e69b7d2f9b47e93dd9cea70b3fe0f2f543a/bblfsh/pyuast.cc#L679)):*
> // we call PyDict_New we returned a new reference

I wrote retruned. RETRUNED :joy: . Corrected now!

ncordon:
___
*[bblfsh/pyuast.cc, line 686 at r9](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll899US5j2MXzGOccur:-LlBnMjM7DvcaG0lDSHA:b-896fix) ([raw file](https://github.com/bblfsh/python-client/blob/23c22e69b7d2f9b47e93dd9cea70b3fe0f2f543a/bblfsh/pyuast.cc#L686)):*

As 679, ibid.
</blockquote></details>

Done.

ncordon:
___
*[bblfsh/pyuast.cc, line 603 at r9](https://reviewable.io/reviews/bblfsh/python-client/176#-Ll86p60B9WR0VNFT6JT:-LlBnSzM3lXoCcLbOeYW:b-896fix) ([raw file](https://github.com/bblfsh/python-client/blob/23c22e69b7d2f9b47e93dd9cea70b3fe0f2f543a/bblfsh/pyuast.cc#L603)):*

Thanks, this comment is really helpful. I have some simplifying suggestions you should feel free to ignore:

```c
// The STL container used to cache the objects does not support concurrent writes, but
// Python's GIL (Global Interpreter Lock) ensures lookupOrCreate and createIfNotExists
// will not execute concurrently, avoiding concurrent writers to the cache.
```
</blockquote></details>

Done.

PR #182, opened 2019-08-02
SupportedLanguages protocol V2 \w aliases.

kuba--:
Signed-off-by: kuba-- <kuba@sourced.tech>
Closes https://github.com/bblfsh/python-client/issues/168

Apart from the extra method `SupportedLanguages` with aliases, this PR get rid of protocol v1 for python-client.

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/182)


bzz:
Commit history and some change look a bit suspicious - is that intentional?

Screen Shot 2019-08-08 at 7 51 29 PM

kuba--:
I've rebased. we can squash during merge or i can rebase and squash again to just one commit

PR #183, opened 2019-08-02
Fix context deallocation

### LINKED ISSUE #181: Avoid context deallocation in Node and NodeIterator
Currently the classes [Node](https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/node.py#L127) and [NodeIterator](https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/node_iterator.py#L9) save a reference to the context to avoid context dellocation. We should take care of that from the C++ side instead, and do a `Py_INCREF` of `ctx` on `Node` creation (and `Py_DECREF` on `Node` deletion) and likewise for iterators.

Right now we have an imbalanced `Py_INCREF` in the filter methods:

https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/pyuast.cc#L333

https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/pyuast.cc#L912

And there is no `Py_INCREF` of the context at `Node` creation:

https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/pyuast.cc#L478

https://github.com/bblfsh/python-client/blob/dfda83d6d29e4a385690deda450c6f077ee1e84c/bblfsh/pyuast.cc#L485

Related issue: #167. This was first tried to fixed in #176 but decided to be treated differently,

ncordon:
- [x] Balance `Py_INCREF`s when possible. In particular the easiest way of balancing here is remove the ones from `filter` methods in the C code, and let Python keep a reference to the context. Certainly it is not the best way to do it, but it involves fewer changes. 
- [x] Open an issue to describe the problems (#185) with the current design of `ResultContext`, `Node` and `NodeIterator` and how to slightly redesign them to be able to move all the reference tracking to the C part.
- [x] Tests for the context, iterators, etc DECREFs (and check their underlying nodes are alive once they go out of scope)

Closes #181

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/183)


ncordon:
*[a discussion](https://reviewable.io/reviews/bblfsh/python-client/183#-LlWTaiMDDJPDsMXndh3:-LlanhE9-sD_P76EHC-0:b-tvz8nq) (no related file):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

Overall looks good, but if that would be possible to improve the "form" i.e make comments describing the tests less verbose/shorter/contain less boilerplate - that would be really help the understanding.

E.g a prefix like `This test is intended to test that ` does not really assist any understanding of the test itself and can be safely skipped.

Another idea - it may make sense to have only one long preamble before all 3 tests, explaining the details (similar to the one each test has right now) and then a very short one before the tests just explaining the difference.
</blockquote></details>

Done Alex!

---

ncordon:
___
*[bblfsh/test.py, line 398 at r1](https://reviewable.io/reviews/bblfsh/python-client/183#-LlIojPS8J4kBTfqUArV:-LlanpOpEhJX8B2g0e0h:by2ler) ([raw file](https://github.com/bblfsh/python-client/blob/460e8dd432289d90d4545e4be9ceb7300163d087/bblfsh/test.py#L398)):*

Yeah, I think @ncordon meant that the value should be released.

For the test failure, it's usually a segfault. But worth documenting, of course.
</blockquote></details>

No no, the value is not released. When you perform a `del` in Python, it `DECREF`s the reference of the underlying `CPython` object  (and you are allowed to do that only once in the Python side). Since an iterator needs the underlying tree to work, even `DECREF`ing the tree from the Python side should not delete the tree until the iterator has gone out of scope.

ncordon:
___
*[bblfsh/test.py, line 400 at r1](https://reviewable.io/reviews/bblfsh/python-client/183#-LlIomWP8LAbQD6TIAKy:-Llaoq-y71VMAbKW4xyH:bsj6465) ([raw file](https://github.com/bblfsh/python-client/blob/460e8dd432289d90d4545e4be9ceb7300163d087/bblfsh/test.py#L400)):*

typo: iterators
</blockquote></details>

Corrected!

ncordon:
___
*[bblfsh/test.py, line 417 at r1](https://reviewable.io/reviews/bblfsh/python-client/183#-LlWsX9j72IW14xeZKqq:-Llaoufr4XEoomlc-1pE:b2qtk2p) ([raw file](https://github.com/bblfsh/python-client/blob/460e8dd432289d90d4545e4be9ceb7300163d087/bblfsh/test.py#L417)):*

Are you sure this will actually deallocate? Maybe worth running a GC afterward?

Have you tested that it actually fails every single time if the ref balance is incorrect?
</blockquote></details>

It does not deallocate the `ctx` in fact (and we do not want it to do so, because otherwise when I try to advance the iterator I find that I do not have the unrderlying tree). As I have explained, this `DECREF`s the count of references to `ctx` (`NodeIterator` still has one since that is what the `fitler` method creates) . And we are only allowed to do `del variable` once in `Python` (by constraint of the interpreter itself). The `ctx` would in fact get released when `it` (or other objects using it) goes out of scope

ncordon:
___
*[bblfsh/test.py, line 417 at r1](https://reviewable.io/reviews/bblfsh/python-client/183#-LlWsX9j72IW14xeZKqq:-LlapmX60Z9nIMO6EzZa:b-36ho4o) ([raw file](https://github.com/bblfsh/python-client/blob/460e8dd432289d90d4545e4be9ceb7300163d087/bblfsh/test.py#L417)):*
> Have you tested that it actually fails every single time if the ref balance is incorrect?

Yes, I tested it when doing #176

ncordon:
___
*[bblfsh/test.py, line 437 at r1](https://reviewable.io/reviews/bblfsh/python-client/183#-LlIqMHp3boDADRXMBcE:-Llaq4HX652XWbT6ojL7:bjwvdct) ([raw file](https://github.com/bblfsh/python-client/blob/460e8dd432289d90d4545e4be9ceb7300163d087/bblfsh/test.py#L437)):*

typo: iterators
</blockquote></details>

Corrected now (I hope!)

ncordon:
___
*[bblfsh/test.py, line 483 at r1](https://reviewable.io/reviews/bblfsh/python-client/183#-LlWUNEQBLMjyZlWiPo1:-Llaq7NYCL7FEcHErMgC:b-oqqk0z) ([raw file](https://github.com/bblfsh/python-client/blob/460e8dd432289d90d4545e4be9ceb7300163d087/bblfsh/test.py#L483)):*
<details><summary><i>Previously, bzz (Alexander) wrote…</i></summary><blockquote>

Is there a reason why we prefer the comments before the function to the python [docstrings](http://google.github.io/styleguide/pyguide.html#381-docstrings)?
</blockquote></details>

Well, now that I have unified the comments for all `testOrphan{whatever}` in a single comment, it does not make sense to use Python docstrings, does it (correct me if I am wrong? Python docstrings are tied to a function (whereas here we have several of them documented with a single comment) so that documentation can automatically be generated for that method (in form of html for example). Here we are not documenting a method of the package, but a test (dev docs), so maybe python docstrings are not necessary?

ncordon:
___
*[bblfsh/test.py, line 398 at r1](https://reviewable.io/reviews/bblfsh/python-client/183#-LlIojPS8J4kBTfqUArV:-Llb4nmD3qInepxlwJx2:b96ia0b) ([raw file](https://github.com/bblfsh/python-client/blob/460e8dd432289d90d4545e4be9ceb7300163d087/bblfsh/test.py#L398)):*

Yeah, sorry, I meant it will be released in case we have a bug.
But I'm now wondering if we need to trigger a GC as well, just in case.
</blockquote></details>

Oh yes, I did not think of that. Good catch Denys. I will add that ASAP

ncordon:
___
*[bblfsh/test.py, line 398 at r1](https://reviewable.io/reviews/bblfsh/python-client/183#-LlIojPS8J4kBTfqUArV:-Llb67mQ6ZV3wm0SgRGT:b-896fja) ([raw file](https://github.com/bblfsh/python-client/blob/460e8dd432289d90d4545e4be9ceb7300163d087/bblfsh/test.py#L398)):*

Oh yes, I did not think of that. Good catch Denys. I will add that ASAP
</blockquote></details>

Done!

ncordon:
___
*[bblfsh/test.py, line 419 at r3](https://reviewable.io/reviews/bblfsh/python-client/183#-LlbJETbEkOjK8Ussa4R:-LlbVQrc-7G4cULvP4et:b-896fja) ([raw file](https://github.com/bblfsh/python-client/blob/d3e33542ecaee30b19a3f85718626b1a20afb26f/bblfsh/test.py#L419)):*

Another optional suggestion: the `del; gc()` pattern now repeats a lot in tests, so it may deserve a separate helper.
</blockquote></details>

Done!

ncordon:
*[a discussion](https://reviewable.io/reviews/bblfsh/python-client/183#-Llbkd9Q-_vm1ufvFNmk:-LlbmJkD45COYyfVNRiX:bg23dez) (no related file):*
I think I did not introduced that, and is failing because Travis cannot download some dependency:

> rpc error: code = Unknown desc = Get https://registry-1.docker.io/v2/bblfsh/cpp-driver/manifests/latest: dial tcp 54.175.43.85:443: i/o timeout

I have tried rebooting the job in Travis, but gives the same error.

---

ncordon:
*[a discussion](https://reviewable.io/reviews/bblfsh/python-client/183#-Llbkd9Q-_vm1ufvFNmk:-Llbr7vDBoyHpsXOfJYX:b-eysjl5) (no related file):*

I think I did not introduced that, and is failing because Travis cannot download some dependency:

> rpc error: code = Unknown desc = Get https://registry-1.docker.io/v2/bblfsh/cpp-driver/manifests/latest: dial tcp 54.175.43.85:443: i/o timeout

I have tried rebooting the job in Travis, but gives the same error.
</blockquote></details>

Now it's working

---

PR #180, opened 2019-07-31
Bumps libuast version

ncordon:
- [x] Updates `TreeOrder` to include `ANY_ORDER` and ~~`CHILDREN_ORDER`~~, now supported by `libuast` (`CHILDREN_ORDER` is not working right now, so we will introduce it in further PRs)
- [x] Drops hard coded `TreeOrder`s and lifts the values directly from C
- [x] Bumps `libuast` version
- [x] Adds tests for new orders

Closes #179 

Signed-off-by: ncordon <nacho.cordon.castillo@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/180)

PR #187, opened 2019-08-07
Adds children order

ncordon:
It needs a new release of  [sdk](https://github.com/bblfsh/sdk) and [libuast](https://github.com/bblfsh/libuast) to work (travis would fail until we do them). It also includes d86fefe which should have been included in #180 :see_no_evil: and adds the order for positions as is in the SDK, instead of lexicographical order of `(offset, line, column)`

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/187)

### DIFF
--- a/bblfsh/test.py
+++ b/bblfsh/test.py
@@ -330,6 +330,14 @@ class BblfshTests(unittest.TestCase):
         self.assertEqual(set(expanded), {'root', 'son1', 'son2', 'son1_1',
                                          'son1_2', 'son2_1', 'son2_2'})
 
+    def testAnyOrder(self) -> None:
+        root = self._itTestTree()
+        it = iterator(root, TreeOrder.CHILDREN_ORDER)
+        self.assertIsNotNone(it)
+        expanded = self._get_nodetypes(it)
+        # We only can test that the order gives us all the nodes
+        self.assertEqual(set(expanded), {'son1', 'son2'})
+
     # Iterating from the root node should give the same result as
     # iterating from the tree, for every available node
     def testNodeIteratorEqualsCtxIterator(self) -> None:
--- a/bblfsh/test.py
+++ b/bblfsh/test.py
@@ -270,16 +270,21 @@ class BblfshTests(unittest.TestCase):
 
     @staticmethod
     def _position_sort(x, y):
-        (a, b, c) = x
-        (e, f, g) = y
+        (off1, line1, col1) = x
+        (off2, line2, col2) = y
 
-        if a == 0 or e == 0 or (b == 0 and c == 0) or (f == 0 and g == 0):
-            if b != c:
-                return b - f
-            else:
-                return c - g
+        def has_offset(off, line, col):
+            off != 0 or (line == 1 and col == 1)
+
+        # Extracted from here
+        # https://github.com/bblfsh/sdk/blob/a2cab92/uast/uast.go#L124
+        if has_offset(off1, line1, col1) and has_offset(off2, line2, col2):
+            return off1 - off2
         else:
-            return a - b
+            if line1 != line2:
+                return line1 - line2
+
+            return col1 - col2
 
     def testIteratorPreOrder(self) -> None:
         root = self._itTestTree()
@@ -336,7 +341,7 @@ class BblfshTests(unittest.TestCase):
         self.assertIsNotNone(it)
         expanded = self._get_nodetypes(it)
         # We only can test that the order gives us all the nodes
-        self.assertEqual(set(expanded), {'son1', 'son2'})
+        self.assertEqual(expanded, ['son1', 'son2'])
 
     # Iterating from the root node should give the same result as
     # iterating from the tree, for every available node
--- a/bblfsh/test.py
+++ b/bblfsh/test.py
@@ -11,6 +11,7 @@ from bblfsh.client import NonUTF8ContentException
 from bblfsh.node import NodeTypedGetException
 from bblfsh.result_context import (Node, NodeIterator, ResultContext)
 from bblfsh.pyuast import uast, decode
+from functools import cmp_to_key
 
 class BblfshTests(unittest.TestCase):
     BBLFSH_SERVER_EXISTED = None
@@ -267,6 +268,19 @@ class BblfshTests(unittest.TestCase):
                                    "start" in x["@pos"].keys(), nodes) ]
         return [ (int(n["offset"]), int(n["line"]), int(n["col"])) for n in start_positions ]
 
+    @staticmethod
+    def _position_sort(x, y):
+        (a, b, c) = x
+        (e, f, g) = y
+
+        if a == 0 or e == 0 or (b == 0 and c == 0) or (f == 0 and g == 0):
+            if b != c:
+                return b - f
+            else:
+                return c - g
+        else:
+            return a - b
+
     def testIteratorPreOrder(self) -> None:
         root = self._itTestTree()
         it = iterator(root, TreeOrder.PRE_ORDER)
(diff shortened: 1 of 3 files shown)


ncordon:
___
*[bblfsh/test.py, line 272 at r1](https://reviewable.io/reviews/bblfsh/python-client/187#-LlgxdzS1t3f5JchekeL:-Llkogty-vXdNYnAbDNE:blw1mex) ([raw file](https://github.com/bblfsh/python-client/blob/3450a20f1ab3e0e123163d8e04127cec85528854/bblfsh/test.py#L272)):*
<details><summary><i>Previously, kuba-- (Kuba Podgórski) wrote…</i></summary><blockquote>

Let me understand, how do you want to sort these tuples?
By (offset, line, col)?
I mainly try to understand the condition: ` a == 0 or e == 0 or (b == 0 and c == 0) or (f == 0 and g == 0):`
</blockquote></details>

Well, for starters this is completely wrong (I do not know how I copied it so wrongly, seriously, I have made a huge mess with the names :joy:). It should be, with `(a, b, c) = (offset, line, column)`:

```
# If any of the positions do not have offset
if (a == 0 or e == 0) and ((b != 1 or c != 1) or (f != 1 and g != 1)):
    if b != f:
        return b - f
    else:
        return c - g
# In other case, compare by offset
else:
    return a - e
```

I have copied this conditions from [here](https://github.com/bblfsh/sdk/blob/a2cab92d9ac4d154241d773e36f2760c6fae134d/uast/uast.go#L124)

ncordon:
___
*[bblfsh/test.py, line 272 at r1](https://reviewable.io/reviews/bblfsh/python-client/187#-LlgxdzS1t3f5JchekeL:-LlktzeB9vTCW-6ZmeJn:b-6s3ijz) ([raw file](https://github.com/bblfsh/python-client/blob/3450a20f1ab3e0e123163d8e04127cec85528854/bblfsh/test.py#L272)):*

Well, for starters this is completely wrong (I do not know how I copied it so wrongly, seriously, I have made a huge mess with the names :joy:). It should be, with `(a, b, c) = (offset, line, column)`:

```
# If any of the positions do not have offset
if (a == 0 or e == 0) and ((b != 1 or c != 1) or (f != 1 and g != 1)):
    if b != f:
        return b - f
    else:
        return c - g
# In other case, compare by offset
else:
    return a - e
```

I have copied this conditions from [here](https://github.com/bblfsh/sdk/blob/a2cab92d9ac4d154241d773e36f2760c6fae134d/uast/uast.go#L124)
</blockquote></details>

I do not know why do we consider that for `line = 1` and `column = 1` the position always has an offset

ncordon:
___
*[bblfsh/tree_order.py, line 12 at r1](https://reviewable.io/reviews/bblfsh/python-client/187#-LlgpvqlBY6zWnm6Ei6T:-LlkuE7q79LD2mIxdK-r:bq70mbd) ([raw file](https://github.com/bblfsh/python-client/blob/3450a20f1ab3e0e123163d8e04127cec85528854/bblfsh/tree_order.py#L12)):*

I've seen you removed it in the previous one, but assumed that was intentional :D
</blockquote></details>

Yes yes, removing it from the previous PR was intentional

ncordon:
___
*[bblfsh/test.py, line 272 at r1](https://reviewable.io/reviews/bblfsh/python-client/187#-LlgqDcN3rYQgaxysKBd:-LlkuQk11Hwkea5T1q2_:b7d0of7) ([raw file](https://github.com/bblfsh/python-client/blob/3450a20f1ab3e0e123163d8e04127cec85528854/bblfsh/test.py#L272)):*

I wound say we can remove this (it's a copy from SDK, right?) and just hardcode the correct order in the test.
</blockquote></details>

But what happens if we decide to write another test using positions? We would have to make it a method in the class again

ncordon:
___
*[bblfsh/test.py, line 272 at r1](https://reviewable.io/reviews/bblfsh/python-client/187#-LlgqDcN3rYQgaxysKBd:-LllRO04EB6hEgCQraPI:bd3oofx) ([raw file](https://github.com/bblfsh/python-client/blob/3450a20f1ab3e0e123163d8e04127cec85528854/bblfsh/test.py#L272)):*

Or just run the test, manually verify that order makes sense and add it to the assertion.
</blockquote></details>

I am okay with that. However I would like to add more extensive tests to the `sdk` or start doing them from now on. IMHO testing against a tree of just 7 or 8 nodes does not guarantee that orders work. These functions should be tested either against:

* Bigger trees generated from a sufficiently large file (I was using `fixtures/test.py` here)
* Random trees (we generate a random tree(s) of bounded size in the tests and we run the same test I was trying to do here.

wdyt?

ncordon:
___
*[bblfsh/test.py, line 272 at r1](https://reviewable.io/reviews/bblfsh/python-client/187#-LlgxdzS1t3f5JchekeL:-LlmFwzD1yIfRvwdYOKT:boqcde4) ([raw file](https://github.com/bblfsh/python-client/blob/3450a20f1ab3e0e123163d8e04127cec85528854/bblfsh/test.py#L272)):*

This is why I don't want to reimplement SDK logic here. There are a few rules there, as you both noticed. Maybe more will be added in the future.
</blockquote></details>

Erased this method from the tests as @dennwc suggested

ncordon:
___
*[bblfsh/test.py, line 272 at r1](https://reviewable.io/reviews/bblfsh/python-client/187#-LlgqDcN3rYQgaxysKBd:-LlmG40PETju9byk1-VY:b-ue7vru) ([raw file](https://github.com/bblfsh/python-client/blob/3450a20f1ab3e0e123163d8e04127cec85528854/bblfsh/test.py#L272)):*

I don't know what you mean by "sufficiently large", but I think ~12 nodes might be enough. Generating an iteration test for a giant tree will make it unverifiable by a human.

Generating a random tree is not an option either. It may be too simple, and tests won't fail even if the iteration order is wrong. Also, this will require the client to copy the implementation of each iterator order which I want to avoid.

But I agree that SDK needs better test coverage for iterators. That's a separate question, though.
</blockquote></details>

I disagree with the statement about the tree with 12 nodes. It does not mean that I do not agree with testing against things a human can check manually. But on top of that I would throw checking of mathematical properties for random or big stuff. For example for the positional order, we could check that in fact the positions I get iterating over the tree in `POSITIONAL_ORDER` are sorted. 

I agree maybe it should be in the `sdk` instead of here, but definitely I do not think it would do harm to add checking of certain properties in addition to the tests that can be manually inspected. For me if ∀x p(x) (for all input x the property x should hold), it makes more sense to check against randomized inputs (because in a computer we cannot just yet test easily that something holds for all inputs). When we are testing against a single tree of 12 nodes, we are missing a lot of edge cases, other sizes, etc. These are my personal opinions of course.

ncordon:
*[a discussion](https://reviewable.io/reviews/bblfsh/python-client/187#-Lln00R81iuVUExkiLJG:-Lln4rFw3k_weCt1Xyuw:b-erjb88) (no related file):*

The CI is still failing for some reason.
</blockquote></details>

Because a release of `sdk` had to be made (you did so today), one of `libuast` (with the new `sdk`) and finally bump the version here

---

ncordon:
*[a discussion](https://reviewable.io/reviews/bblfsh/python-client/187#-Lln00R81iuVUExkiLJG:-Lm9iL_2A_p7HqtGnMcp:b-iylco) (no related file):*

Can we update it in this PR to make sure the CI passes?
</blockquote></details>

We need a new release of `libuast` using the `sdk v3.2.3` to do so (right now `libuast` is using the version `v3.2.0`). There is no rush to do so. I have bumped the `sdk`version [here](https://github.com/bblfsh/libuast/pull/109)

---

ncordon:
*[a discussion](https://reviewable.io/reviews/bblfsh/python-client/187#-Lln00R81iuVUExkiLJG:-LmG8FwS6qGysTRIPTqF:b-8q1vig) (no related file):*

New libuast release: https://github.com/bblfsh/libuast/releases/tag/v3.4.2
</blockquote></details>

Updated and CI passed :)

---

bzz:
woot, I'm sorry for merging - was just checking the GitHub status and it seems to be able to merge it now alright 😆

PR #188, opened 2019-08-07
Adds Python context functionality

### LINKED ISSUE #173: Allow creating contexts from Python objects
It would be useful if we could get the python dict from a node, change it and then reload the node with the change to use .filter().

Something like:

```
context = client.parse(filename)
modified_dict = rename_variables(context.root.get())
context.load(modified_dict)
```

ncordon:
I have wrapped the functionality in a `Context` class (`iterate`, `filter`, `encode`) and provided tests for it. It allows to have a Python memory context and work with it.

Tests may fail for now because `encode` is transforming `False`s into `0`s and `decode` is not inverting that behavior:

```
from bblfsh import *

ctx = ResultContext()
obj = {"k1": False, "k2": "v2"}
fmt = 1 # YAML

encoded = ctx.ctx.encode(obj, fmt)
encoded
# bytearray(b'{\n   \'k1\': 0,\n   \'k2\': "v2",\n}')
decoded = decode(encoded, fmt)
decoded.load()
#  {'k1': 0, 'k2': 'v2'}
```

Closes #173

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/188)


ncordon:
___
*[bblfsh/context.py, line 6 at r1](https://reviewable.io/reviews/bblfsh/python-client/188#-LlhMIEREL4TzX0x9K6T:-LlklFiwCjE3a28vP7Am:bsrsxqn) ([raw file](https://github.com/bblfsh/python-client/blob/4a30edc47f357ce2dc927b4ca0b583026e0cebd1/bblfsh/context.py#L6)):*

Does it make sense to partially join this type with a `ResultContext`?
</blockquote></details>

What do you mean by partial join? `ResultContext` includes even more functionality (loading all the nodes for example and storing the grpc response). What we could definitely do is to include the`encode` method in `ResultContext` also, because right now if we have a `ctx : ResultContext` we have to call `ctx.ctx.encode(None, 0)` to encode it (correct me if I am wrong)

ncordon:
___
*[bblfsh/context.py, line 6 at r1](https://reviewable.io/reviews/bblfsh/python-client/188#-LlhMIEREL4TzX0x9K6T:-Lo0Hnd90CHvKxsuAPg9:butps2k) ([raw file](https://github.com/bblfsh/python-client/blob/4a30edc47f357ce2dc927b4ca0b583026e0cebd1/bblfsh/context.py#L6)):*

I'm just wondering if `ResultContext` may embed `Context`.

You are right, `Context` shouldn't contain a response. And `load` doesn't makes sense either. But everything else is similar, right?

As for the `encode`, yes I think it may be useful for the `ResultContext` as well. For example, encoding into YAML after decoding from the binary format (e.g. gRPC).
</blockquote></details>

Well I think we should keep them separated, since in the `scala-client` they are (and they should be because of the structure of JNI we are using)

ncordon:
*[a discussion](https://reviewable.io/reviews/bblfsh/python-client/188#-LlhMQkGDPxNgUMkXRgI:-Lo0I2zE6288vLPuiplI:b-8g677) (no related file):*

Blocking because of the failing test.
</blockquote></details>

This is ready to review now

---

ncordon:
Done @kuba-- @

PR #190, opened 2019-09-11
Fixes exception when passing incorrect bytes to decode

### LINKED ISSUE #189: Incorrect bytes in decode kill the python interpreter
Thanks to @smacker who discovered this issue:

```python
from bblfsh import *
bytes = b''
decode(bytes, format = 0)
```
kills the Python intrepreter. Also any incorrect sequence of bytes like `bytes = b'abcdef' would produce the same behavior.

ncordon:
An uncaught exception was thrown and python interpreter was killed because of that.

Closes #189 

Signed-off-by: ncordon <nacho.cordon.castillo@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/190)


ncordon:
___
*[bblfsh/pyuast.cc, line 1060 at r1](https://reviewable.io/reviews/bblfsh/python-client/190#-LoWJy_UDD5QxkHo6fpq:-LoWNzy214AFpdIPHWeH:b-c5x97t) ([raw file](https://github.com/bblfsh/python-client/blob/86ddb9546a012ef2b6e616ffd987dacdd409f85c/bblfsh/pyuast.cc#L1060)):*

In this case we haven't released `buf`. Should we fall through as you did in the case below?
</blockquote></details>

I think `buf` would be released after the `try-catch`, or am I missing something?

ncordon:
___
*[bblfsh/pyuast.cc, line 1060 at r1](https://reviewable.io/reviews/bblfsh/python-client/190#-LoWJy_UDD5QxkHo6fpq:-LoWS-uHDoBycgVcugLK:bjza2wd) ([raw file](https://github.com/bblfsh/python-client/blob/86ddb9546a012ef2b6e616ffd987dacdd409f85c/bblfsh/pyuast.cc#L1060)):*

I don't see how—if we `return nullptr` at this point control exits the function and that code never runs, right?
</blockquote></details>

Oh yes I missed that, my bad. Fixing it

ncordon:
___
*[bblfsh/pyuast.cc, line 5 at r2](https://reviewable.io/reviews/bblfsh/python-client/190#-Lo_Gkw0ERNhdeDvSiAs:-LoeII0tEvZU3VhIGSJ5:blmaq4x) ([raw file](https://github.com/bblfsh/python-client/blob/7f039ce7bf60b9f6c8479b98ef9ad53383bc069f/bblfsh/pyuast.cc#L5)):*

I was also wondering why it's there, but forgot to ask :)
</blockquote></details>

I erased it

PR #194, opened 2019-10-01
Adds timeout of 60 seconds to parse requests

ncordon:
Mirrors https://github.com/bblfsh/scala-client/pull/133. We already had a `timeout` parameter here, but with default value of `None` (no timeout). Default timeout has been set to 60 seconds as in `scala-client`

Signed-off-by: ncordon <nacho.cordon.castillo@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/bblfsh/python-client/194)

### DIFF
--- a/bblfsh/client.py
+++ b/bblfsh/client.py
@@ -55,7 +55,7 @@ class BblfshClient:
 
     def parse(self, filename: str, language: Optional[str]=None,
               contents: Optional[str]=None, mode: Optional[ModeType]=None,
-              timeout: Optional[int]=None) -> ResultContext:
+              timeout: int=60) -> ResultContext:
         """
         Queries the Babelfish server and receives the UAST response for the specified
         file.
@@ -68,13 +68,16 @@ class BblfshClient:
         :param contents: The contents of the file. IF None, it is read from \
                          filename.
         :param mode:     UAST transformation mode.
-        :param timeout: The request timeout in seconds.
+        :param timeout: The request timeout in seconds. Zero or negative \
+                        means no timeout.
         :type filename: str
         :type language: str
         :type contents: str
         :type timeout: float
         :return: UAST object.
         """
+        if timeout is None or timeout <= 0:
+                timeout = None
 
         # TODO: handle syntax errors
         contents = self._get_contents(contents, filename)


bblfsh/client.py
```diff
@@ -55,7 +55,7 @@ def _get_contents(contents: Optional[Union[str, bytes]], filename: str) -> str:
 
     def parse(self, filename: str, language: Optional[str]=None,
               contents: Optional[str]=None, mode: Optional[ModeType]=None,
-              timeout: Optional[int]=None) -> ResultContext:
+              timeout: int=60) -> ResultContext:
```
kuba--: so after you removed an `Optional` what `0` means? 
Is it "no timeout - wait forever"?

ncordon: `None` was wait forever in fact. I have tested the `0` and negative values and produces a `DEADLINE_EXCEEDED`:

``` 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.7/site-packages/bblfsh/client.py", line 84, in parse
    response = self._stub_v2.Parse(request, timeout=timeout)
  File "/usr/lib/python3.7/site-packages/grpc/_channel.py", line 565, in __call__
    return _end_unary_response_blocking(state, call, False, None)
  File "/usr/lib/python3.7/site-packages/grpc/_channel.py", line 467, in _end_unary_response_blocking
    raise _Rendezvous(state, None, None, deadline)
grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with:
	status = StatusCode.DEADLINE_EXCEEDED
	details = "Deadline Exceeded"
```

ncordon: If you mean if there is still a way of waiting forever, passing `None` to the method should suffice. I am adding a comment to make it clear

kuba--: yeah usually 0 or negative number means - wait forever.
In our case, we had an `Option::None` and every `Some(value)` was just a regular timeout.
So, my point is, if we want to switch from `Optional` to some primitive type, like int, long, then 0 should stand for `infinite`, because most likely does not make sense to pass 0ms timeout.

ncordon: Done!

ncordon:
___
*[bblfsh/client.py, line 80 at r3](https://reviewable.io/reviews/bblfsh/python-client/194#-LqHQOQvCyfA9_rtlRIs:-LqHT5PG2xm7xfXplnUD:bj2d02g) ([raw file](https://github.com/bblfsh/python-client/blob/e5bcce89045bfb3f55b237454bf1be34847c1fb1/bblfsh/client.py#L80)):*

I don't think we need an explicit type check here—we've declared that the parameter is an `int` so it should be sufficient to just compare the value.

```python
if timeout <= 0:
   timeout = None
```


</blockquote></details>

Yes, but python does no type checking, so I can still pass whatever I want as timeout (e.g. a string)

ncordon:
___
*[bblfsh/client.py, line 80 at r3](https://reviewable.io/reviews/bblfsh/python-client/194#-LqHQOQvCyfA9_rtlRIs:-LqHTGMtAxHZBzfsv66o:b-pzh022) ([raw file](https://github.com/bblfsh/python-client/blob/e5bcce89045bfb3f55b237454bf1be34847c1fb1/bblfsh/client.py#L80)):*

Yes, but python does no type checking, so I can still pass whatever I want as timeout (e.g. a string)
</blockquote></details>

And since before parameter was an `Optional[int]` I could pass a `None` (no timeout) and I want to preserve that behaviour to not break the API (in case someone was passing `None` explicitly)

ncordon:
___
*[bblfsh/client.py, line 72 at r3](https://reviewable.io/reviews/bblfsh/python-client/194#-LqHQp-S9yLC6LGupPPQ:-LqHTV813d11JZoV8qW6:bdmezmc) ([raw file](https://github.com/bblfsh/python-client/blob/e5bcce89045bfb3f55b237454bf1be34847c1fb1/bblfsh/client.py#L72)):*

Also zero: `Zero or negative means no timeout.` perhaps

Since the default value is already written in the function header, I don't think we need to repeat it here—it'll only get out of sync if we update it later.
</blockquote></details>

True, changed!

ncordon:
___
*[bblfsh/client.py, line 80 at r3](https://reviewable.io/reviews/bblfsh/python-client/194#-LqHQOQvCyfA9_rtlRIs:-LqHZ2fc8YHqZBUUR_gR:b-tnd1z0) ([raw file](https://github.com/bblfsh/python-client/blob/e5bcce89045bfb3f55b237454bf1be34847c1fb1/bblfsh/client.py#L80)):*

Of course it doesn't check, as you say—but we also don't explicitly check any other arguments either. It's normal and idiomatic in Python to assume you have a value that matches the documented type, and allow the program to fail at runtime in the usual way (e.g., in the comparison operator). It seems odd to add a type check for this one value and not the others.

Moreover, our explicit check masks what we would otherwise see as an error, if someone passes in some arbitrary blob of data. 
</blockquote></details>

Yes, but again, if I do not that, if someome was relying on the API as `parse(..., timeout = None)` if I remove that type checking I would break that call

ncordon:
___
*[bblfsh/client.py, line 80 at r3](https://reviewable.io/reviews/bblfsh/python-client/194#-LqHQOQvCyfA9_rtlRIs:-LqHjmSf6EYu9CZHP6I4:b-cq6kl8) ([raw file](https://github.com/bblfsh/python-client/blob/e5bcce89045bfb3f55b237454bf1be34847c1fb1/bblfsh/client.py#L80)):*

In that case, how about just handling that specific case?

```python
if timeout is None or timeout <= 0:
   timeout = None
```

That way we don't silently ignore someone passing (say) a string, but still deal with the existing uses?
</blockquote></details>

Definitely better, yes. Changed!

